PR #105, opened 2022-01-21
added bug template

RishiKumarRay:


andrewtavis (approved):
Looks great! I'm going to merge and take a look at it live so that I can see the length/flow, but this is exactly what I was hoping for. Will make some edits and get back to you in the issue when I'm done with them :)

Thanks for your efforts, @RishiKumarRay! 😊😊

Issue #132, opened 2022-02-12
Add Italian keyboard


andrewtavis:
### Terms

- [X] I have searched [open new keyboard issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3A%22new+keyboard%22)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] This issue is about one language, and I've changed the title to reflect this

### Language support

Italian grammar functions similarly to other latin languages, so the formatting scripts for adding the data should largely be able to be replicated from those. Nouns are either masculine or feminine, so again this will be similar.

### Contribution

I'll be working on this for v1.3.0, but any help would be much appreciated! 😊 

### What's needed

- [x] An Italian directory added to [Keyboards/LanguageKeyboards](https://github.com/scribe-org/Scribe-iOS/tree/main/Keyboards/LanguageKeyboards) that includes an `ItalianKeyboardViewController.swift` file
- [x] Italian language keyboard values and logic added to `ITInterfaceVariables.swift` in the above directory
- [x] Italian language keyboard commands added to `ITCommandVariables.swift` in the above directory
- [ ] Fine tuning the WDQS queries found in [Scribe-iOS/Data/Italian](https://github.com/scribe-org/Scribe-iOS/tree/main/Data/Italian)
- [ ] Writing formatting Python scripts and adding them to [Scribe-iOS/Data/Italian](https://github.com/scribe-org/Scribe-iOS/tree/main/Data/Italian)
- [x] Add Italian keyboard specific logic to [KeyboardViewController.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/KeyboardViewController.swift)
- [x] Setting up alternate key functions if there are keys that have them that aren't already implemented
- [ ] Adding all the formatted JSON data to a `Data` directory in the new directory and updating the data table in the readme

andrewtavis:
Word counts via [Data/check_language_data.sparql](https://github.com/scribe-org/Scribe-iOS/blob/main/Data/check_language_data.sparql):

type | total
:-- | --:
adjective | 60
adverb | 15
noun | 562
personal pronoun | 2
preposition | 20
pronoun | 6
proper noun | 6
verb | 70

andrewtavis:
[...]

PR #137, opened 2022-02-15
#132 add baseline Italian keyboard without working Scribe commands

andrewtavis:


andrewtavis:
To now is a baseline Italian keyboard without working Scribe commands. See issue for progress via the checkboxes :)

Issue #26, opened 2021-11-23
Hold-to-select design


andrewtavis:
This issue is extremely important. Scribe needs to create and implement a proper [Bezier path](https://developer.apple.com/documentation/uikit/uibezierpath) or other design for hold-to-select characters, with this issue including keys popping up when being pressed and removing key pressed coloration where no longer needed. 

At time of writing hold-to-select characters are the most glaring design flaw for Scribe as they simply do not look like the UI for system keyboards. As this is very unexpected behavior and also is not properly implemented, any help with designs and implementation would be much appreciated!

One feature of the current design that should be maintained, if only as an option when #16 is finished, is that the alternate character view persists in Scribe, whereas in system keyboards the user needs to continue to hold and drag to select.

andrewtavis:
Now that the key designs have been updated, it should be much easier to derive exact key locations and create elements to show a proper key pop up. This will not be implemented for Scribe command buttons as feedback indicates that it's unnecessary, and further their location at the top of the keyboard prevents this feature as Scribe elements cannot exceed the keyboard height.

PR #140, opened 2022-02-22
#26 add key popup and redesign hold-to-select chars

andrewtavis:


andrewtavis:
Currently what's needed is to set up the pop up views for keys that are on the left and right of the keyboard so that they're fully within the keyboard view. After that the hold-to-select keys need to be changed so that they use a similar view to the pop ups.

andrewtavis:
Only hold-to-select callouts need to be edited at this point :)

andrewtavis:
From here there’s just the need to check that it’s all being sized properly across devices and clean the code a bit so that more of the consistent spacing arithmetic is saved in variables.

Issue #168, opened 2022-05-30
Switch conjugation back to base view


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

This is a very simple issue. Currently the conjugation state is not resetting for the conjugation command, which means that if a user conjugated a verb in past tense, then the next time they use the command past tense will be the first one shown. Rather what we'd like is for conjugation to always start with present tense, or whatever the most simple conjugation forms are.

What needs to happen is that the conjugation state needs to be reset after the command is executed, which could go in the block for `case "return"` of `executeKeyActions` of [KeyboardsBase/KeyboardViewController.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/KeyboardViewController.swift).

### Device type

iPhone and iPad

### iOS version

I assume this is happening for all versions.

andrewtavis:
This seems to only be affecting certain keyboards and not others. The German keyboard does switch back, for instance, but the Spanish one does not.

andrewtavis:
@SaurabhJamadagni, this issue deals with resetting the conjugation state before the interface is loaded. Looking at it, I only did it for German and not any of the other ones (I'm normally testing on the German keyboard). The conjugation state is reset for German at this place in the codes, which is right when the conjugation button is pressed:

https://github.com/scribe-org/Scribe-iOS/blob/076650f8cd1f36c811636a226d4dbe83340705eb/Keyboards/KeyboardsBase/KeyboardViewController.swift#L1272

The line `deConjugationState = .indicativePresent` resets german as German's base conjugation is this one as defined below:

https://github.com/scribe-org/Scribe-iOS/blob/076650f8cd1f36c811636a226d4dbe83340705eb/Keyboards/LanguageKeyboards/German/DECommandVariables.swift#L23

Basically we want to do the same for the baseline conjugation states for the other languages, which are all found in the corresponding `CommandVariables.swift` files. Six lines are needed - one for each missing conjugation reset - and maybe a space afterwards would also help with readability :)

SaurabhJamadagni:
Thank you very much @andrewtavis! I'll get started with this immediately.

[...]

PR #172, opened 2022-06-28
Fix: Returning conjugation to base conjugation for all keyboard languages

SaurabhJamadagni:
- PR for issue #168 
- Added required conjugation code references to [KeyboardsBase/KeyboardViewController.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/KeyboardViewController.swift) for all keyboard languages.


andrewtavis (approved):
Exactly what we're looking for! Let me bring it in and then test to make sure that this fixes the bug 🐞

Thank you!

Issue #35, opened 2021-11-23
Make command bar cursor more apparent


andrewtavis:
This issue is to add a minor UI element to Scribe keyboards where the command bar would be shown to be actionable when `commandState == true` in [KeyboardViewController.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/KeyboardViewController.swift) and [CommandVariables.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/CommandVariables.swift). As of now the cursor and prompt are static and in one color, but adding a blink or greyed out "write" in the keyboard language would make it more apparent to the user that they will be typing in the command bar instead.

Changing the prompt to have greyed out text would likely be the easiest solution to this (see image below).

- [x] This issue should also include updating the `Translate` images in the App Store to show the command placeholder

andrewtavis:
![SCRIBE_3](https://user-images.githubusercontent.com/24387426/158158627-2a8a91bd-c9e2-43e3-9b09-1c18a36413b0.png)

The above was done by a designer as an example for a blogpost header. This could be used as a guide for doing the greyed out text, which could be `keyPressedColor` from [ColorVariables.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/ColorVariables.swift).

andrewtavis:
We could also do "Scribe" in greyed out letters to keep it consistent across keyboards.

SaurabhJamadagni:
Hi @andrewtavis, I would like to help with this task. I am new to iOS Development. It would be really helpful if you could give me a few pointers as to what needs to be done. For the greyed out letters, I was wondering if adding a placeholder text property to `commandBar` is the solution?

andrewtavis:
@SaurabhJamadagni, thanks for reaching out! I'd be very happy to give some direction on this. Your suggestion to add a placeholder text is definitely in the right direction, but then this placeholder text would only be viewable when `commandState == true` (we want to keep the UI clean when the user is not in command mode). Big thing as well is that we need to make sure that when the user starts typing the "write" prompt goes away, but then if they delete back to the start of the prompt it needs to return.

[...]

PR #175, opened 2022-07-10
Feat: Adding placeholder prompts to Command Bar

SaurabhJamadagni:
- PR for second stage of issue #35 
- Added `conditionallyRemovePlaceholder()` and `conditionallyPlacePlaceholder()` to [CommandBar.swift](https://github.com/scribe-org/Scribe-iOS/blob/6200cbcd8457187b815b67e8316d99261737f944/Keyboards/KeyboardsBase/ScribeFunctionality/CommandBar.swift).
- These allow us to remove the placeholder prompt when the user starts typing and re-insert it when the text is deleted.
- Added PromptAndPlaceholder variables to `allPrompts` array to stop the prompt from being deleted.
- Added call to appropriate function for the `default`, `spaceBar`, `"'"`, `delete` and `keyLongPressed` cases.

### DIFF
--- a/Keyboards/KeyboardsBase/InterfaceVariables.swift
+++ b/Keyboards/KeyboardsBase/InterfaceVariables.swift
@@ -118,7 +118,7 @@ func setKeyboardLayout() {
     setLayoutFxn()
   }
 
-  allPrompts = [translatePromptAndCursor, conjugatePromptAndCursor, pluralPromptAndCursor]
+  allPrompts = [translatePromptAndCursor, conjugatePromptAndCursor, pluralPromptAndCursor, translatePromptAndPlaceholder, conjugatePromptAndPlaceholder, pluralPromptAndPlaceholder]
 }
 
 // Variables that define which keys are positioned on the very left, right or in the center of the keyboard.
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -1091,7 +1091,7 @@ class KeyboardViewController: UIInputViewController {
       // Always start in letters with a new keyboard.
       keyboardState = .letters
       loadKeys()
-      commandBar.text = translatePromptAndCursor
+      commandBar.text = translatePromptAndPlaceholder
 
     // Switch to conjugate state.
     case "Conjugate":
@@ -1099,7 +1099,7 @@ class KeyboardViewController: UIInputViewController {
       commandState = true
       getConjugation = true
       loadKeys()
-      commandBar.text = conjugatePromptAndCursor
+      commandBar.text = conjugatePromptAndPlaceholder
 
     // Switch to plural state.
     case "Plural":
@@ -1112,7 +1112,7 @@ class KeyboardViewController: UIInputViewController {
       commandState = true
       getPlural = true
       loadKeys()
-      commandBar.text = pluralPromptAndCursor
+      commandBar.text = pluralPromptAndPlaceholder
 
     // Move displayed conjugations to the left in order if able.
     case "shiftConjugateLeft":
@@ -1193,6 +1193,8 @@ class KeyboardViewController: UIInputViewController {
       loadKeys()
 
     case "delete":
+      // Inserting the placeholder when commandBar text is deleted
+      commandBar.conditionallyPlacePlaceholder()
       if shiftButtonState == .shift {
         shiftButtonState = .normal
         loadKeys()
@@ -1214,6 +1216,7 @@ class KeyboardViewController: UIInputViewController {
       clearCommandBar()
 
     case spaceBar:
+      commandBar.conditionallyRemovePlaceholder()
       if commandState != true {
         proxy.insertText(" ")
         if [". ", "? ", "! "].contains(proxy.documentContextBeforeInput?.suffix(2)) {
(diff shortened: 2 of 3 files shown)


SaurabhJamadagni:
@andrewtavis please review these changes. I look forward for your feedback. Thank you 😊

andrewtavis (changes requested):
This is all really great, @SaurabhJamadagni! Honestly I was telling my GF earlier today how happy I am to have such dedicated contributions, as this by no means is a small endeavor :)

I'm unsure if you would prefer to makes changes yourself or have me do them, but here are a few minor things I'm realizing should be done for this PR:

- All the instances of `translatePromptAndPlaceholder` and the like should be defined as `translatePromptAndCursor + " " + translatePlaceholder` (my bad for not suggesting this, but am realizing there should be a space between the prompt and the placeholder).
- I think that `conditionallyPlacePlaceholder` is a bit repetitive, so maybe `conditionallyAddPlaceholder`? 😊
- `conditionallyRemovePlaceholder` needs to be after the delete action in the `case` block, as the way it's working now is that I need to delete back and then press delete one more time for it to recognize that we're at the start of the prompt, whereas it should be that delete is pressed and then the conditions are checked to add the placeholder if need be.
- For translate, the prompt has the target language as the source language as well now (so it's `de -› de` instead of `en -› de`, but this is also only when the user has yet to type anything). I checked and this is happening because I didn't direct you to also make changes to the English interface variables within [KeyboardsBase/InterfaceVariables.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/InterfaceVariables.swift). Within there we also need a `translatePromptAndPlaceholder` that will be accessed such that we get `en -› de`. We can also add the other placeholder variables for the first stage of this to `setENKeyboardLayout`. Hopefully this will fix that, but then there might be another minor change that needs to happen if it's still `de -› de` 😊

Let me know if you want to do another commit, or if I should make the above changes. Happy either way, and thanks so much for your efforts on this!

SaurabhJamadagni:
> Honestly I was telling my GF earlier today how happy I am to have such dedicated contributions, as this by no means is a small endeavor :)

That is so sweet @andrewtavis! I am glad I can be of help :)

> All the instances of translatePromptAndPlaceholder and the like should be defined as translatePromptAndCursor + " " + translatePlaceholder

Yes! I did observe that and I could tell something didn't feel right. I just couldn't tell what was missing though. Thanks for pointing it out.

> Let me know if you want to do another commit, or if I should make the above changes. Happy either way, and thanks so much for your efforts on this!

Yes I would love to makes the changes. It is absolutely no problem. I take it as my responsibility to see this problem through from start to finish. I would be more than happy to make the changes you have requested. I will get started ASAP. 😊

andrewtavis:
Looking forward to the next commit, @SaurabhJamadagni! :)

andrewtavis:
> * `conditionallyRemovePlaceholder` needs to be after the delete action in the `case` block

@SaurabhJamadagni, reading this again, it should be `conditionallyAddPlaceholder` For this point :)

SaurabhJamadagni:
@andrewtavis, I have implemented your suggestions. Please review the changes. Thank you :)

SaurabhJamadagni:
@andrewtavis, I was studying a few search bars and I noticed that most that have a prompt don't have a space between the cursor and the text. For example, the apple spotlight search bar and the search bar in firefox as references respectively.
image
image
Do you think maybe when we grey out the text it will look better, or should we keep the space? Do let me know your thoughts. I just wanted to point out an observation. I hope you don't mind 😊

andrewtavis:
> @andrewtavis, I have implemented your suggestions. Please review the changes. Thank you :)

Getting to it shortly 😊 Thank you!

> Do you think maybe when we grey out the text it will look better, or should we keep the space? Do let me know your thoughts. I just wanted to point out an observation. I hope you don't mind 😊

I appreciate the initiative in thinking about this! Really great :) Let's keep the spaces for now and check how it looks, then remove them if it looks weird. Thing is that I'm not sure if we're going to be doing the blinking cursor for the command bar (at least Spotlight's blinks, not sure on Firefox), and if we don't then that means that we'd inherently have a different design spec that wouldn't necessarily be the same as the UI you've mentioned. If the cursor is blinking, then there are moments where the text stands by itself, but if not then there likely should be a space.

Reason that I don't think we'll do the blink for the command bar cursor is that in most cases there already will be a blinking cursor - the cursor that's in the actual text field. Open to your feedback on this, but the thought of two blinking cursors on screen at the same time does not sound appealing, especially when you consider that sometimes they'll be in sync and other times they won't 😅 Assuming we don't have the command bar cursor blink, I'm expecting that the final command prompt with the placeholder will look better if there's a space, but then let's of course check this assumption.

SaurabhJamadagni:
> Open to your feedback on this, but the thought of two blinking cursors on screen at the same time does not sound appealing, especially when you consider that sometimes they'll be in sync and other times they won't 😅 Assuming we don't have the command bar cursor blink, I'm expecting that the final command prompt with the placeholder will look better if there's a space, but then let's of course check this assumption.

Yes that's a fair assumption. Let's wait for the end result. I also like the space, but I was just curious as to if 'no space' is a standard. But it's a completely valid point that we can have our own design choices. Let's go ahead with the version you have suggested. I really like the look of it. It feels very balanced.

andrewtavis (approved):
All looks great!! Thanks so much for getting to this, @SaurabhJamadagni! 😊

Issue #178, opened 2022-07-21
Switch URL in App Screen to iOS Repo


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This is a very simple issue. Currently the link to GitHub in the app screen goes to the organization, but instead it should go to this repository. This change would happen in [ViewController.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Scribe/ViewController.swift) and [ENAppText.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Scribe/AppTexts/English/ENAppText.swift).

Reason for this is that most people who click the link will be mostly curious about Scribe-iOS.

### Contribution

I'd be happy to get to this if no one else has interest :)

andrewtavis:
@SaurabhJamadagni, messaging you in here so that I can hopefully assign you. Wasn’t letting me select you, but maybe if you’re linked in the chat it’ll work :)

andrewtavis:
Hmmmmm, still not. Could you confirm in here that you want to work on this, and I’ll assign it? Also, just checking, does this change make sense to you?

SaurabhJamadagni:
> Could you confirm in here that you want to work on this, and I’ll assign it? Also, just checking, does this change make sense to you?

Yes, @andrewtavis I would like to work on this one. It makes sense to route the users coming from the iOS app to the iOS repo and respectively for the android counterpart :)

andrewtavis:
Thanks for your willingness to help, and further for the feedback that we should do this 😊

andrewtavis:
Thank you for the PR, @SaurabhJamadagni! I'm thinking post e0a6b85 above that you're welcome to edit the [CHANGELOG.md](https://github.com/scribe-org/Scribe-iOS/blob/main/CHANGELOG.md) to describe your work in future PRs :) Obviously I'm happy to do it myself, so only if you'd like to 😊

SaurabhJamadagni:
[...]

PR #179, opened 2022-07-23
Fix: Changing Scribe-org link to Scribe-iOS repo link

SaurabhJamadagni:
- PR for issue #178 
- Changed URL to direct to Scribe-iOS repository in files `ViewController.swift` and `ENAppText.swift` instead of the Scribe-org page.

@andrewtavis, please review these changes. Thank you :)


andrewtavis (approved):
Thanks for the quick edit, @SaurabhJamadagni! I actually like that the text in-app wasn't changed so that it's still shorter, but just the link takes them to the iOS repo :)

Issue #165, opened 2022-05-18
Change return key icon to a right triangle during command executions


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This is a very simple issue. The goal is that the commands Translate, Conjugate or Plural would trigger a change in the icon that's applied to the return key. Not only would it change color to `ScribeBlue`, but it would also change its icon to `arrowtriangle.right.fill` of SF Symbols from `arrow.turn.down.left`, with the original icon being replaced once a command is canceled or executed. 

General feedback has been that it's a bit hard to realize that the return key should be pressed to execute commands, so adding this might make it more apparent.

- #103 could potentially he changed such that `arrowtriangle.right.fill` or `arrow.right` could be used when the user is making a web search.
- This could similarly include switching the button to `magnifyingglass` if the user is searching.

### Contribution

Would be happy to do this if others don't have interest :)

SaurabhJamadagni:
> As of now I'm thinking that arrowtriangle.right.fill might be best as it would be very apparent for new users, as the rest of the UI is similar to arrowtriangle.right, which thus might blend in too much.

@andrewtavis, the `arrowTriangle.right` does seem like a good option. The return key changes colour when the command state is true. I think that coupled with the `arrowTriangle.right.fill` symbol should make it prominent enough.

andrewtavis:
Glad you agree :) Looking forward to this change! We can also include it in v1.4.0 if you plan to get to it quickly, but no stress if you need more time. I’ll likely try to release by early next week at the latest. 

Also, note that maybe the size of the icon will need to change as some of them look a bit off at certain dimensions. Hopefully this isn’t the case, but I can give more directions if you think `arrowTriangle.right.fill` looks too small or big.

SaurabhJamadagni:
> We can also include it in v1.4.0 if you plan to get to it quickly, but no stress if you need more time.

Yeah sure, I will get to it at the earliest. Fingers crossed, this should be straightforward and shouldn't take long 😅

I'll let you know if I need help. Thanks!

andrewtavis:
[...]

PR #180, opened 2022-07-23
UI: Changing return key symbol to arrowTriangle.right.fill

SaurabhJamadagni:
- PR for issue #165 
- Changed `arrow.turn.down.left` to` arrowTriangle.right.fill` in files [AppTextStyling.swift](https://github.com/scribe-org/Scribe-iOS/blob/bddbaaa1805d244f852ac46ec07eb47e4b253c0a/Scribe/AppTexts/AppTextStyling.swift) and [KeyboardViewController.swift](https://github.com/scribe-org/Scribe-iOS/blob/bddbaaa1805d244f852ac46ec07eb47e4b253c0a/Keyboards/KeyboardsBase/KeyboardViewController.swift)
- Added required symbol to [keysThatAreSlightlyLarger](https://github.com/scribe-org/Scribe-iOS/blob/bddbaaa1805d244f852ac46ec07eb47e4b253c0a/Keyboards/KeyboardsBase/KeyboardStyling.swift#L32) array to increase the size of the symbol relative to the other keys.

@andrewtavis, please review these changes at your convenience. Thank you so much 😊

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardStyling.swift
+++ b/Keyboards/KeyboardsBase/KeyboardStyling.swift
@@ -35,7 +35,8 @@ var keysThatAreSlightlyLarger: [String] = [
   "chevron.right",
   "shift",
   "shift.fill",
-  "capslock.fill"
+  "capslock.fill",
+  "arrowtriangle.right.fill"
 ]
 
 /// Get the icon configurations for keys if the device is an iPhone.
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -880,7 +880,7 @@ class KeyboardViewController: UIInputViewController {
           }
 
           if key == "return" {
-            styleIconBtn(btn: btn, color: keyCharColor, iconName: "arrowtriangle.right.fill")
+              styleIconBtn(btn: btn, color: keyCharColor, iconName: commandState ? "arrowtriangle.right.fill" : "arrow.turn.down.left")
           }
 
           if key == "delete" {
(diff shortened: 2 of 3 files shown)


andrewtavis (changes requested):
I feel that the issue wasn't explained properly, @SaurabhJamadagni, and for that I apologize 😞 This issue is to switch the icon only when a command is being used, where `arrow.turn.down.left` will then be changed to `arrowTriangle.right.fill` at the same time that the return key is turned Scribe blue. The changes within the app screen of `arrow.turn.down.right` are also unnecessary and should be reverted. 

Where this change should occur is within the `if key == "return"` block  of [KeyboardViewController.swift](Keyboards/KeyboardsBase/KeyboardViewController.swift) as you derived, but rather than always, we could do a boolean check of `if commandState == true` where `arrowTriangle.right.fill` would be passed to `styleIconBtn`, and the else clause would instead pass `arrow.turn.down.left`.

On the positive side, you're right there and this all should be very easy to fix 😊 Most importantly you did find where the changes should be made, and the icons will now be appropriately sized because of our conversation in #180.

Again, my apologies that the issue wasn't as well formulated as it could be!

SaurabhJamadagni:
> I feel that the issue wasn't explained properly, @SaurabhJamadagni, and for that I apologize 😞 This issue is to switch the icon only when a command is being used, where arrow.turn.down.left will then be changed to arrowTriangle.right.fill at the same time that the return key is turned Scribe blue. The changes within the app screen of arrow.turn.down.right are also unnecessary and should be reverted.

Oh, I understand now. I am very sorry! I must have misread a few lines in the excitement of solving it. I will revert the unnecessary changes right away. 😅

SaurabhJamadagni:
@andrewtavis, I have implemented the corrections. I used a ternary operator instead of `if-else` as I thought that the code would look more uniform as the other keys such as `shift` and `delete` don't have any substantial body inside the parenthesis. I could change it to an if-else if you would like.

andrewtavis (approved):
Thanks for this! I'm realizing that there's a mini bug in this that's not in the scope of this, which was my fault in the original implementation. I'll bring this in and make a new issue for that 😊 Thanks for getting to the edits so quickly!

Issue #188, opened 2022-08-09
Alphabetized Autocomplete


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to add an initial auto-suggest feature to Scribe, and is the first part of #3. Work for this will be done in relation to #187 for the [Wikimania Hackathon](https://wikimania.wikimedia.org/wiki/Hackathon).

This issue comprises:
- [x] Finding resources for a function to find the next words alphabetically
- [x] Figuring out which words should be suggested to the user
    - At the start, just nouns
- [x] Removing the command bar unless the keyboard is in command mode
    - Annotations need to show up after a user presses space still
    - The coloration likely needs to change to `keyboardBgColor`, as making it clear would prevent annotations from being presented
- [x] Fixing the coloration of characters on annotations to match the new background
- [x] Properly styling the Scribe key corners/shadow when the command bar is removed
- Putting two lines between the command buttons to delineate them when they have no background (@andrewtavis)
    - [x] Add the lines to the .xib file
    - [x] Create functions that allow them to be added and removed
    - [x] Removing the left line if an annotation of long enough length is being shown
        - This needs to be done in the function to show the lines
        - Checking the length of the text in the command bar is the likely method for this
- [x] Make the command buttons user interactable when an autocomplete would be possible
    - The buttons work now, and are conditionally deactivated if an annotation is happening based on the length of the annotation
- [x] The creation of a function to find the three next words in alphabetical order to the currently typed word in the proxy
- [x] Assigning those three words to the command buttons
- [x] Running the function and doing the assignment each time a key is pressed (within `loadKeys()`)
- [x] Inserting the word on a button to the text field when a user presses an auto suggest button

### Contribution

I'll be working on this with @SaurabhJamadagni :)

andrewtavis:
@SaurabhJamadagni, I added something to the above, as we need to remember that even though the command bar won't be on the screen always, we're still going to need to use it to do noun and preposition annotations. Here's again the picture of eventual coloration in autocomplete:

https://user-images.githubusercontent.com/24387426/182542168-f8e337f4-4709-413c-86e4-e6ded470eed7.png

[...]

PR #189, opened 2022-08-12
#188 hackathon: add alphabetical autocomplete

andrewtavis:
First part of #188 - I need to go to bed 😴😅

This PR includes everything that's ticked off in the issue. I'm generally referring to things as "auto actions", as we'd use the same functionality for autocomplete and autosuggest.

We now have autocomplete fields that lack functioning buttons. The fields are made by making the command bar the same color as the keyboard background, and then I added two little lines (`partitions`) to the .xib that show up when we're not in command mode or conjugating. When annotation of noun genders occurs, then the partition between fields 1 and 2 disappears if the word + annotation is long enough (I can explain this better). 

I'm having issues getting the buttons to effectively be clickable, and specifically we want to disable the first and maybe second autocomplete buttons if the user is being shown a gender annotation. Ex: user types `Leben` on a German Scribe keyboard, which has two annotations and is thus long enough to go into the second field. We thus remove the partition to make one field, but then we want the first and second buttons to then not be visible or work. I'm trying to do this with conditions via `autoAction1Visible` and `autoAction2Visible` when annotations are happening, but it's not working 🤔

If you want to see the current attempt, then uncomment `// conditionallySetAutoActionBtns()` in the `KeyboardViewController.swift`. Maybe check it with it commented so you can see the new UI 😊

Random changes:
- Switched `keyboardBackColor` to `keyboardBgColor`

I'll be up early, and maybe we can get on a call? Big thing you can do is try to plan out how to do the alphabetical selection and start working on the function. We can potentially just put the function in `KeyboardViewController.swift`, as it will need to be controller the command buttons directly.

### DIFF
--- a/Keyboards/KeyboardsBase/ColorVariables.swift
+++ b/Keyboards/KeyboardsBase/ColorVariables.swift
@@ -19,7 +19,7 @@ var commandKeyColor = UIColor(red: 255.0/255.0, green: 255.0/255.0, blue: 255.0/
 var commandBarColor = UIColor(red: 255.0/255.0, green: 255.0/255.0, blue: 255.0/255.0, alpha: 1.0)
 var commandBarBorderColor = UIColor(red: 255.0/255.0, green: 255.0/255.0, blue: 255.0/255.0, alpha: 1.0).cgColor
 
-var keyboardBackColor = UIColor(red: 255.0/255.0, green: 255.0/255.0, blue: 255.0/255.0, alpha: 1.0)
+var keyboardBgColor = UIColor(red: 255.0/255.0, green: 255.0/255.0, blue: 255.0/255.0, alpha: 1.0)
 var keyShadowColor = UIColor(red: 255.0/255.0, green: 255.0/255.0, blue: 255.0/255.0, alpha: 1.0).cgColor
 
 // annotate colors.
@@ -47,7 +47,7 @@ func checkDarkModeSetColors() {
     commandBarColor = UIColor.commandBarColorLight
     commandBarBorderColor = UIColor.commandBarBorderColorLight
 
-    keyboardBackColor = UIColor.keyboardBackColorLight
+    keyboardBgColor = UIColor.keyboardBgColorLight
     keyShadowColor = UIColor.keyShadowColorLight
 
     annotateRed = UIColor.annotateRedLight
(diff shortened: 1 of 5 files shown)


SaurabhJamadagni:
> I'll be up early, and maybe we can get on a call? Big thing you can do is try to plan out how to do the alphabetical selection and start working on the function.

Can we do a call after 09:30 UTC @andrewtavis? I have a family gathering going on right now and there's a lot of people entering my room frequently 😂 I'll get started on the function immediately.

I just had a trivial doubt. How can I run the changes on a simulator? Do I have to checkout into the PR and pull the changes or something? I am sorry, I have never done this before 😅

andrewtavis:
Morning ☀️ I'll likely be busy for a moment at 9:30 UTC, but no stress if we can't do a call :) We can maybe do one in the afternoon before. I'll be in touch. 

Your doubt's not trivial :) Everyone needs to learn this stuff something 😊 There are some nice instructions online for how to checkout branches, but I think that GitHub also has some on this page. Right now as a reviewer you should have `view command line instructions` to checkout the branch at the bottom of the page. This [SO question](https://stackoverflow.com/questions/1783405/how-do-i-check-out-a-remote-git-branch) also has some nice details :)

andrewtavis:
@SaurabhJamadagni, proof as always that sleep and breaks are needed, as I figured out what I was having issues with 😅😊

[443b2da](https://github.com/scribe-org/Scribe-iOS/pull/189/commits/443b2da7c986aa996a39ca91aab0cc3c42a78246) makes the buttons conditionally active based on if an annotation is being shown or not. The third button - the `Plural` one - is always active. The first button is deactivated if there's an annotation, and the second button is deactivated if the annotation is long enough that it would interfere with the display of the annotation 🎉

SaurabhJamadagni:
> There are some nice instructions online for how to checkout branches, but I think that GitHub also has some on this page. Right now as a reviewer you should have view command line instructions to checkout the branch at the bottom of the page. This [SO question](https://stackoverflow.com/questions/1783405/how-do-i-check-out-a-remote-git-branch) also has some nice details :)

Cool! Thanks @andrewtavis! No worries about a call in the afternoon whenever you are free. I am available anytime afterwards. 😊

SaurabhJamadagni:
@andrewtavis, I was successfully able to checkout the branch and pull the changes and the change in UI looks really great! I will make my changes right away. I am just in awe with how much you were able to get done haha. Anyway, I'll get back to the task. Talk to you soon! :)

SaurabhJamadagni (approved):
Understood! 😊

PR #191, opened 2022-08-13
Adding functions for autocomplete and assigning buttons

SaurabhJamadagni:
- PR for second part issue #188 
- Added the required function that generate the next three words alphabetically for autocomplete.
- Set the buttons with the the appropriate text based on this array.
- This function takes into account only one word at a time. 
  - That is when you enter a word and move to the next word in the sentence, it will suggest autocomplete only for that word.
  - Also replaces the typed word with the auto completion instead of just adding it to the end of the user input.

@andrewtavis could you please review these changes! It was so nice to walk through the problem with you. No way was I otherwise figuring it out 😂

Thanks for the help!

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -210,6 +210,63 @@ class KeyboardViewController: UIInputViewController {
 
     removeLeftAutoActionPartition = false // reset partition removal
   }
+  
+  /// Generates the array for the three autocomplete words.
+  func getAutocompleteWords() {
+    let keysOfNouns = Array(nouns!.keys).sorted()
+    completionWords = [" ", " ", " "]
+    if proxy.documentContextBeforeInput?.count != 0 {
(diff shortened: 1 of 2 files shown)


andrewtavis:
@SaurabhJamadagni, I didn't understand the process of the PRs and merged the other one, but then now the changes here are on a branch that's deleted, so it apparently was automatically closed.....

Sorry 🤦‍♂️

Can you resend this?

andrewtavis:
Never mind, I figured it out :) I have autodelete for branches setup for this repo, so it got deleted, but I just needed to recheckout my local branch and push again 😊

andrewtavis (approved):
All looks great, and thanks :) Thanks for all the hard work! I'll play around with it a bit and we can maybe edit it a bit so that there's always three words being shown if the buttons are available. I'll also likely go through and just make a dedicated sorted array that lacks words with hyphens. I'm wondering if getting the keys and sorting them each time you run the function is one of the reasons why it's all a bit slow 🤔

We can chat this all over a bit more in #188 😊 Great work!! 👏

Issue #194, opened 2022-08-14
Baseline autosuggest


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In order to release the work that's been done in #188, Scribe also needs to add baseline autosuggest functionality. Reason being is that the user would expect to also see word suggestions even if they're not currently typing a word. This is the other part of the baseline version of #3, which would see this feature implemented using natural language processing models over the current words in the text proxy (will be another issue later).

The working idea for this is that we should find/generate a list of most used words in each Scribe keyboard language, and then do random selections from this list. In this way the architecture of the feature will be set for future iterations.

### Contribution

I'll be working on this with those who are interested prior to the v1.5.0 release :)

andrewtavis:
@SaurabhJamadagni, for this issue, how does it sounds if we use [wordfreq](https://github.com/rspeer/wordfreq/)? It has a function `top_n_list()` that would give us the words we need, and we could maybe just subset so that we're selecting words that are three characters or longer? I was thinking we can take a look at the words over three characters given certain cutoffs for `n`, and from there we can decide how many words we want to include for this baseline (i.e. 30, 40, etc).

It has an MIT license, so we should mention it in the privacy policy and I'll maybe make a NOTICE.txt for us to add some things that we're using.

Seems like the easiest thing to do, as I can just make a simple Python script and save it to [Scribe-Data](https://github.com/scribe-org/Scribe-Data) :)

andrewtavis:
Looking at German autosuggest for the first time in a bit, some patterns are coming that we could also use to make this more performant, as I think that some of the system autosuggest isn't really NLP based at all it seems. The following two are ones I think we can implement, and maybe we can think of one or two others:

- There are three words that are always presented for the first options when nothing is typed that we can copy
- If the first word the user types is a pronoun, the three suggestions are normally the conjugations of have, be and can

andrewtavis:
[...]

PR #195, opened 2022-08-14
Fixing gaps in autocomplete

SaurabhJamadagni:
- PR for part of issue #188 
- Fixed no capitalisation issue and added cases for CapsLock.
- Fixed autocomplete to work even when delete is pressed.
  - This works by having a secondary variable called `secondaryPastStringOnDelete` that stores the previous version of the text in proxy. That way we can return the `pastStringInTextProxy` to the previous line.

@andrewtavis, could you please review these changes. Also could you look into the bug that crashes the keyboard when `delete` is pressed consecutively when the text proxy is empty? I mentioned the exact nature of it in the main issue comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+See the [releases for Scribe-iOS](https://github.com/scribe-org/Scribe-iOS/releases) for an up to date list of versions and their release dates. Versions that are marked as released may yet to be in the App Store if it's within the 48 hour submission review period.
+
 Scribe-iOS tries to follow [semantic versioning](https://semver.org/), a MAJOR.MINOR.PATCH version where increments are made of the:
 
 - MAJOR version when we make incompatible API changes
@@ -12,8 +14,8 @@ Emojis for the following are chosen based on [gitmoji](https://gitmoji.dev/).
 
 ### ✨ New Features
 
-- Scribe now includes a baseline autocomplete feature that suggests the next possible noun based on the current word being entered.
-- Scribe now includes a baseline autosuggest feature that suggests words from a list of most common words in the language.
+<!-- - Scribe now includes a baseline autocomplete feature that suggests the next possible noun based on the current word being entered.
+- Scribe now includes a baseline autosuggest feature that suggests words from a list of most common words in the language. -->
 
 ### 🗃️ Data Added
 
--- a/Keyboards/KeyboardsBase/Extensions.swift
+++ b/Keyboards/KeyboardsBase/Extensions.swift
@@ -60,6 +60,10 @@ extension String {
       $1 == char ? $0 + 1 : $0
     }
   }
+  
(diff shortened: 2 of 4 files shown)


andrewtavis (changes requested):
This is all really good, and I'll definitely use this in the showcase 😊 We'll figure out the delete button press issues later. Long story short, pressing delete has definitely been a problem for various features in the past. What I often need to do is just cancel the actions of the delete button under certain conditions, but then that's not ideal.

Reason for requested changes is that we still need to work a bit on capitalizing the words. If I'm in a language like Spanish and I start typing a with shift at the start, then eventually the autocomplete suggestions will change to lower case - this even though I'm at the start of the proxy. I think that maybe a boolean is needed for this, whereby if shift is pressed, then we're in capitalized mode until such a time as space is pressed or another action that would lower case the words takes place. Let's do some testing in the next week to figure it all out 😊

SaurabhJamadagni:
> Reason for requested changes is that we still need to work a bit on capitalizing the words. If I'm in a language like Spanish and I start typing a with shift at the start, then eventually the autocomplete suggestions will change to lower case

I see the problem with it @andrewtavis. The words in the JSON are itself lowercased. So even when starting with capitalized letters we are getting lowercased completions. Having a boolean sounds like a good idea. Again a wrong assumption on my part that all words in the JSON would be capitalized. Yes, let's work on it in the coming week :) 

Thanks for the feedback!

SaurabhJamadagni:
Hi @andrewtavis, I chose against adding one more boolean variable because we still haven't figured out why the keyboard crashes on pressing delete continuously. I extended the String class to capitalize our string correctly. I hope that is okay. Please do review the changes 😊

andrewtavis:
Sounds great, @SaurabhJamadagni! I'll get to it when I can. I have a mini weekend trip coming up, but I will have some time no doubt 😊

SaurabhJamadagni:
> I'll get to it when I can. I have a mini weekend trip coming up, but I will have some time no doubt 😊

Yeah absolutely! Have a great trip 😊

andrewtavis:
I’ll get to the review of this tomorrow odds are, @SaurabhJamadagni :) Just FYI 😊

andrewtavis (approved):
All looks good on my end, @SaurabhJamadagni :) I'm going to remove the cancelation of the conjugation button for long annotations right now, and then we can work to cut off words and add `...` so they fit only in the translation space later on.

One final change that I'll do here is to make sure that words that are capitalized are only presented as such. So a German noun shouldn't be presented at lower case even if the user is typing in lower case, as odds are it's not the word that they'll select, but it still looks wrong. I'll do this as I think about maybe expanding the autosuggest words, but then in thinking about it it might make sense to simply add the words from the most common list that we'll find in #194 that don't already appear in the list. I think that all nouns + 500 most common words is a solid baseline for this, but then let's not stress about it as we're already making so much progress 😊

Thanks again so much for your work on this!

PR #200, opened 2022-09-03
#188 refactoring incl commandState to enum

andrewtavis:
@SaurabhJamadagni, soooooooo here this is 😊

I hope this makes our lives much easier :) You'll see that there are more deletions than additions in this one 🚀 

There are lots of little changes in here:
- I renamed `isAlreadyPluralMessage` to `alreadyPluralMsg`, so disregard all those files
- Changes `commandState` to an `enum` as we discussed
- Updated the codes to work with the new version of it
    - This included a ton of work in the main view controller
    - I reordered lots of things so that we're accounting for all the variations of `commandState`

What this fixes:
- Hopefully the road ahead is a lot easier :)
- Fixes issues where the annotations were showing up on top of the invalid message (so translate something like `fdsjkal`, or use the plural command on something that already is plural)
    - Trying to fix this and finally realizing how much I hate all the booleans was the source of all this :) 
- Changes the interface for invalid messages so that it's the command bar and Scribe key next to each other (just like it used to be 🥲)
    - Reason for this is that I figured the proper user flow is that they get an invalid message, they see it, and they can either start typing - which resets the keyboard - or they can jump into the command again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,8 @@ Emojis for the following are chosen based on [gitmoji](https://gitmoji.dev/).
 
 ### ✨ New Features
 
-<!-- - Scribe now includes a baseline autocomplete feature that suggests the next possible noun based on the current word being entered as well as conjugations of common verbs.
-- Scribe now includes a baseline autosuggest feature that suggests words words derived from Wikipedia as following a given word. -->
+<!-- - Scribe now includes a baseline autocomplete feature that suggests the next possible noun based on the current word being entered as well as the most common words in the keyboard language.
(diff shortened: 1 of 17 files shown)


SaurabhJamadagni (approved):
Wow @andrewtavis! The changes look great. They are surely going to make the process here on much easier. I could see the effect already while this review. The code is easier to read and to understand what exactly is happening or not happening during the command state. Thank you for making these changes. You should have really made a sub task out of this. It was a lot of work and I would have loved to help you out haha 😊

andrewtavis:
Ya I think a sub task would have made sense too, @SaurabhJamadagni 😅 Started working on the invalid bug and just into it 😊 Thanks for looking it over, and glad to do all this so it's easier for us!

PR #203, opened 2022-09-04
#188 #194 autocomplete bug fix and autosuggest placeholders

andrewtavis:
Alright, @SaurabhJamadagni, this is a lot more minor improvements to fix the bug where autocomplete wasn't being triggered after a command was used. I got carried away a bit again though 😊

What's included:
- The aforementioned bugfix
- Added an emum `AutoActionState` that determines if we're completing or suggesting
    - I made a boolean at first and was like, "Waaaaaait a minute..... 😅"
- Fixed the sorting of the words for autocomplete so that it's done regardless of case
- Fixed the conditionals within autocomplete as we need to be able to type a lowercase word but get an uppercase/caps suggestion such as `germany -> Germany`
- Added variables for the first three words that appear whenever you load a system version of one of the Scribe languages that can serve as our default autosuggestions (replaced your function with them as you directed)
- Set up a trigger within `conditionallySetAutoActionBtns` to determine which kind of action is being done
- Went through and defined what type of action should follow each kind of key press (spaces -> autosuggest, normal keys -> autocomplete, etc)

Let me know what you think on all this! Just am trying to get the work load down a bit for the both of us. Would be great if you could check out the last bugs in #188 related to deletion :) After that it's on to autosuggest 😊

### DIFF
--- a/Keyboards/KeyboardsBase/Extensions.swift
+++ b/Keyboards/KeyboardsBase/Extensions.swift
@@ -55,13 +55,17 @@ extension String {
     return self == self.uppercased()
   }
 
+  var isCaptalized: Bool {
+    return self == prefix(1).uppercased() + self.lowercased().dropFirst()
+  }
+
   func count(of char: Character) -> Int {
     return reduce(0) {
       $1 == char ? $0 + 1 : $0
     }
   }
   
-  func capitalizingFirstLetter() -> String {
+  func capitalize() -> String {
     return prefix(1).uppercased() + self.lowercased().dropFirst()
   }
 }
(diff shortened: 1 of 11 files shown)


SaurabhJamadagni (approved):
Hey @andrewtavis! The changes look great! That was a good dodge with the enum. Do not wish to spiral into booleans again haha. Thanks for requesting a review. Learned a few tricks by looking at the way you refactor code and use closures and navigate the codebase. Thanks 😊

SaurabhJamadagni:
> Would be great if you could check out the last bugs in https://github.com/scribe-org/Scribe-iOS/issues/188 related to deletion :) After that it's on to autosuggest

Yes absolutely! On it 😊

PR #206, opened 2022-09-10
#188 fix to delete crash + minor template change

### LINKED ISSUE #188: Alphabetized Autocomplete

andrewtavis:
@SaurabhJamadagni, I think this is what we need to close #188 😊 I think that most of the problems were coming from `changePastTextsFromProxy`. The command prompt crash on delete was happening because we weren't checking to make sure that we weren't in a command within this function. Beyond this, I saw that the crash when the cursor is at the start of the proxy also stopped happening when I removed this function and it's call within `handleDeleteButtonPressed`. It seems that everything is working without this function, so could you just check all this and see if I'm removing it by accident? Autocomplete is updating as the user deletes, so maybe something in how the new `CommandState` enum is functioning fixed a bug that made this function not necessary anymore?

### DIFF
--- a/.github/ISSUE_TEMPLATE/data_wikidata.yml
+++ b/.github/ISSUE_TEMPLATE/data_wikidata.yml
@@ -19,9 +19,9 @@ body:
     validations:
       required: true
   - type: textarea
-    id: choice
+    id: description
     attributes:
-      label: Issue
+      label: Description
       placeholder: Let us know what your idea is!
     validations:
       required: true
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -284,27 +284,6 @@ class KeyboardViewController: UIInputViewController {
       }
     }
   }
-  
-  /// Changes variables pastStringInTextProxy and secondaryPastStringOnDelete.
-  /// Note: allows for autocomplete to work when delete is pressed by reversing the changes in the above variables.
-  func changePastTextsFromProxy() {
-    if proxy.documentContextBeforeInput?.count == 0 {
-      pastStringInTextProxy = ""
-      secondaryPastStringOnDelete = ""
-    } else {
-      let stringInProxy = proxy.documentContextBeforeInput!
-      if pastStringInTextProxy == stringInProxy {
-        pastStringInTextProxy = secondaryPastStringOnDelete
-        
-        if !secondaryPastStringOnDelete.isEmpty {
-          secondaryPastStringOnDelete.removeLast()
-        }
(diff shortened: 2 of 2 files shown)


SaurabhJamadagni (approved):
This is a very interesting outcome @andrewtavis! I think what was happening is that when we were checking for text in proxy, if the proxy contained `nil`, it was most likely entering the else case where the variables were being assigned and that would lead to it trying to delete letters that were `nil`. That's my guess. Glad it's fixed though :)

All the functions seem to be working so I think `changePastTextFromProxy` was redundant. I never thought of deleting the function entirely. Maybe that's because I wrote it and I might have been a bit biased. Not the best trait for a programmer haha. Once I figure out debugging, I might recreate this case and try to see how the control was being passed to identify the reason. Thanks for the review. Simply excited that this was solved 😊

andrewtavis:
That's what I'm thinking as well, @SaurabhJamadagni! And no stress on wanting to keep something you wrote! Just look at me and the damn booleans 😅 Glad to have this closed 😊

PR #207, opened 2022-09-12
Adding Autosuggestions function

SaurabhJamadagni:
- PR for issue #194 
- Adds body to `getAutosuggestions` function.
- Shifts suggestions a place to the left if annotations are active.

Hey @andrewtavis! Could you please review these changes. The function may look like it has repetitive code inside it but without it, it was not accepting input in different cases. For example, if the key "ich" was typed as "Ich", we could see that the key is present in the dictionary by comparing the lowercased strings. But what would happen is that when retrieving elements, "Ich" would be a wrong key as the correct key is "ich".

I think the two different cases I have considered work with all the permutations that are possible -- key lowercase and input uppercase OR key uppercase and input uppercase and so on. I hope that works. If you find a way I could simplify this further let me know. I would be glad to make the changes :)

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -282,10 +282,44 @@ class KeyboardViewController: UIInputViewController {
     ]
     
     let prefix = pastStringInTextProxy.replacingOccurrences(of: secondaryPastStringOnDelete, with: "").replacingOccurrences(of: " ", with: "")
-    if dummySuggestions.keys.contains(prefix) {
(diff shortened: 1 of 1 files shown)


andrewtavis:
Awesome, @SaurabhJamadagni! I'll look into this shortly 😊 Maybe there's a way to simplify it, but let's not stress too much about some repeated code if it's getting the job done :)

andrewtavis (approved):
Looks good to me! I'm going to add the autosuggestions after numbers now, and also make autosuggest work after the double space period shortcut, which as of now is not working for some reason 🤔

Thanks for all your efforts, @SaurabhJamadagni!

PR #213, opened 2022-10-01
Add command bar button

misteu:
### What does it do?
- adds method for setting up an info button on `CommandBar`
- adds property `isShowingInfoButton` on `CommandBar` that hides or shows the info button
- implements use of `isShowingInfoButton` for when `commandState == .invalid` and `invalidCommandMsg` is shown (as discussed)
- adds property `infoButtonTapHandler` that is triggered when the new info button is tapped.

### Screenshot
grafik

I'm open to any feedback!

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -47,9 +47,6 @@ class KeyboardViewController: UIInputViewController {
     // Override prior command states from previous sessions.
     commandState = .idle
 
-    // Set height for Scribe command functionality.
-    annotationHeight = nounAnnotation1.frame.size.height
-
     loadKeys()
 
     // Set tap handler for info button on CommandBar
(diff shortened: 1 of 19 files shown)


Keyboards/KeyboardsBase/KeyboardViewController.swift
```diff
@@ -51,6 +51,11 @@ class KeyboardViewController: UIInputViewController {
     annotationHeight = nounAnnotation1.frame.size.height
 
     loadKeys()
+
+    // Set tap handler for info button on CommandBar
+    commandBar.infoButtonTapHandler = {
+      print("Tapped Info Button!")
```
misteu, line 57: This is obviously only an example as we did not discuss what exactly you want to trigger with the button. :)

andrewtavis:
@misteu, this got a bit messy with the conflict resolution as the changes for the other issue appear to be deleted. I'll figure it out. Just wanted to bring it to your attention :)

andrewtavis (approved):
Thanks for this as well, @misteu! Really above and beyond seeing as how the original issue needed to be changed because of the change in the UI for v2.0 😊 I'll make an issue now for adding an action to the button when it's clicked.

Issue #210, opened 2022-09-30
Allow for case conjugation on case annotation button press


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As work on #197 progresses, an opportunity that opens up is the conjugation of case pronouns for the user when they press the annotation of a preposition. We could use the conjugation view for this, with the pronouns in question again being the label of the cells, but instead of verb conjugations it would be what the pronoun becomes in the given case. If a preposition has multiple cases, then in a similar way it would be great to allow the user to scroll through them as with verb conjugations.

Whether this should be included in v2.0.0 is up for discussion, and this would be problematic as as of now there would be no action on noun annotation button presses, but then we can also disable nouns or just leave them actionless and hopefully develop an action for them later 😊 Issues of user understanding should not necessarily hinder the rollout of what would be another very unique and helpful feature 🚀 

### Contribution

I'd be happy to work on this! It's a bit tough of an issue for newcomers, but I'm also happy to support if someone has interest :)

andrewtavis:
We'd need to create variables that have the conversions from prepositions to the versions within the case to get this done, but this is common information that can be gotten from countless sources.

andrewtavis:
This is now going to be included in v2.0.0 so that this functionality can be shown in the App Store videos :)

andrewtavis:
@SaurabhJamadagni, if you could quickly check the PR for this, that'd be great 😊 This is a strong WIP implementation for this. Obviously it's not good that something like "den/die/das" would be inserted into the proxy for the user, as the ideality would be that we'd ask them which of the three they need and insert only that one, but this is what we can do without reworking the conjugation display in a crazy way. Worst comes to worst they just cancel and type it themselves, but having this as a reference is really convenient and something that I think people will really like 🚀

SaurabhJamadagni:
> if you could quickly check the PR for this, that'd be great

Hey @andrewtavis, sorry I couldn't get to this PR today. I'll give it a look tomorrow first thing in the morning. Hope that's okay. Sorry for the delay. Could you give me a few test words or phrases that I can use to check what this implementation brings?

andrewtavis:
[...]

PR #215, opened 2022-10-02
#210 add case annotation display functionality

andrewtavis:
Hey @SaurabhJamadagni! 😊

Firstly, I didn't do the conference, so don't think that I can get to all this while doing something like that at the same time 😄

What this is:
- Disables the annotations for nouns as they don't do anything (can't be clicked and don't have shadows)
- Switches the keyboard to the conjugation display when a preposition annotation is pressed
- Allows for migration through the cases via the left and right buttons in the display
- Makes the first case shown to the users the first one in the annotation
    - All cases will be able to be seen from the display with the left and right keys though
- Adds variables for German and Russian case conjugation displays
- Renaming variables :)


andrewtavis:
[efccc13](https://github.com/scribe-org/Scribe-iOS/pull/215/commits/efccc131aef8275f8a9ddf9db460dbaf77473a6e) is just to add the autosuggest files for the other issue so that I can have them available on the current branch for testing as I try to generate them :)

andrewtavis:
@SaurabhJamadagni, [40c094f](https://github.com/scribe-org/Scribe-iOS/pull/215/commits/40c094f6dbfb3883997538985d39f44de0af9fe2) is likely gonna be the dumbest PR you ever get from me, and might be because it's late, but easter eggs are fun, so why not? 🐣😊 Let me know what you think 😄

Along with that are some formatting fixes to the ranges as you suggested, as well as adding `"Scribe"` to the nouns so that it shows up in auto actions :)

Keyboards/LanguageKeyboards/French/Data/autosuggestions.json
```diff
@@ -0,0 +1 @@
+
```
SaurabhJamadagni, line 1: Just nit picking but extra unnecessary whitespace 😅

SaurabhJamadagni (approved):
Hey @andrewtavis! This is such a cool feature! Definitely doesn't feel like a WIP implementation. Also, loved the Scribe easter egg! Looks good to me and didn't experience any bugs during the limited testing I did. Thanks for the review! This was a fun feature to check out. Has me excited 😊

Keyboards/LanguageKeyboards/French/Data/autosuggestions.json
andrewtavis: I'm honestly really excited too 😊 This is gonna be so great :) :)

Thanks for checking it all!

Issue #219, opened 2022-10-04
Refactor conjugation state changes to use case-switch


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This is a simple issue that would help refactor the code a bit :) I noticed that in the `CommandVariables.swift` files I used `if` statements in the `ConjugationStateLeft` and `ConjugationStateRight` functions. An example is [DECommandVariables.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/LanguageKeyboards/German/DECommandVariables.swift), where I correctly used `switch deCaseConjugationState` in a new issue, but in the `else` clause of `if commandState == .selectCaseConjugation` we still have `if deConjugationState == .indicativePresent`. This should be switched to a `case-switch` statement like the following for `deConjugationStateRight`:

```swift
switch deConjugationState {
case .indicativePresent:
  deConjugationState = .indicativePreterite
case .indicativePreterite:
  deConjugationState = .indicativePerfect
case .indicativePerfect:
  return
}
```

This issue would be to switch all `ConjugationStateLeft` and `ConjugationStateRight` functions across all languages to use `switch-case` statements.

I'd be happy to support someone on this if there's interest! 😊

japanese-goblinn:
@andrewtavis Hey! I would like to implement this refactoring

japanese-goblinn:
@andrewtavis working on it!

andrewtavis:
Hey, @japanese-goblinn! Thanks for your willingness to help, and also for the PR! I'll check this later today 😊

andrewtavis:
Thanks for the efforts here, @japanese-goblinn! Very nice to work with you on this 😊 Closing this as we're all done 🚀

PR #220, opened 2022-10-05
Refactor conjugation state changes to use case-switch #219

japanese-goblinn:
Hey! This is a pull request for #219 issue. I'm used `return` statement is assignment to be consistent with `return` in cases with no state changes.

Also I saw that whole project is using `2 spaces` for indentation so I change `4` to `2` in `.xproj`. Now if you have `4 spaces` in you global Xcode settings when you will work on this project `2 spaces` will be used instead.


andrewtavis:
Thanks for this, @japanese-goblinn! And also for the changes to the project settings. Hard to figure out what they should be sometimes with so many options 😊 Will get to this later :)

andrewtavis (changes requested):
Quick note @japanese-goblinn, looks like you forgot French :) I'll go through and check the conditions when that's added in 😊 

Also, thinking about it, I'm not sure if doing `return` in the cases where we’re switching state is really necessary. I get that it's more consistent with returning when we’re not changing, but all that's happening is an assignment of a global enum instance, whereas having return makes me think that a variable is being assigned.

Would it make more sense on your end to do `break` after the condition that doesn't change the state? We could do that instead? :)

andrewtavis:
Note that I edited the above as I didn’t explain it properly at first :)

japanese-goblinn:
@andrewtavis I see. I'll fix it

japanese-goblinn:
@andrewtavis well I found the reason why I missed French - due to typo in file name. Fixed it too

andrewtavis:
You're fixing all kinds of typos, @japanese-goblinn 😊 Thanks!

andrewtavis:
Will review this shortly :) :)

andrewtavis:
This [SO question](https://stackoverflow.com/questions/24142297/using-break-statement-in-switch) seems to indicate that `break` as we chose is the way to go with switch-case if you don't want it to do anything, btw. Just wanted to check for my own learning 😇 A comment says use if-else instead, but I think that we've figured the best way for this situation, as if the enum values change later then an error will be thrown in those case statements to let us know to change them 😊

japanese-goblinn:
@andrewtavis amazing. looking forward for more contributions 😉

Issue #211, opened 2022-09-30
Disable and visually disable conjugation scroll buttons when applicable


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to functionally and visually disable the buttons that allow the user to scroll through conjugations when pressing them will not change the tense. This could also be used or case preposition conjugation in #210 :)

We'd want the icon of the button to be a bit more grey rather than black for light mode and white for dark mode, and within the conjugation assignment there should be check that would prevent the activation of one of the go left button or go right button if the user is at the first or last tense respectively.

### Contribution

Would be happy to work on this, or support if someone has interest in contributing!

japanese-goblinn:
I think I want to take a look on this one

andrewtavis:
Great! Before I assign it, let me give you a quick rundown of what this might entail and then you can confirm if you want to do it :)

Specifically what we're looking at is the following: https://github.com/scribe-org/Scribe-iOS/blob/561c2367d6a7b679747e8fb1dafc0cfbf1886c52/Keyboards/KeyboardsBase/KeyboardViewController.swift#L581

In that function you'll see where `conjugateShiftLeft` and `conjugateShiftRight` are activated, so we'll need that to be conditional, with the same conditions triggering coloration of the icons for the buttons. The button coloration is done here: https://github.com/scribe-org/Scribe-iOS/blob/561c2367d6a7b679747e8fb1dafc0cfbf1886c52/Keyboards/KeyboardsBase/KeyboardViewController.swift#L1195

You'll see the instances of `styleIconBtn` for `conjugateShiftLeft` and `conjugateShiftRight` below, with us needing to set `keyCharColor` to what I'd argue should be `specialKeyColor` so that it matches the conjugation labels which are an appropriate shade of grey given the mode.

Best way to do this is another enum? Likely better than all my booleans that are still flying around the code? 

```swift
enum ConjViewControls {
  case bothActive
  case leftInactive
  case rightInactive
}
```

[...]

PR #230, opened 2022-10-13
Disable Conjugation View Shift buttons conditionally

SaurabhJamadagni:
- PR for issue #211 
- Added an enum that holds the state for which button should be active.
- Assigned values to variable `conjViewShiftButtonsState` accordingly for all languages in their respective command variables file.
- Changed colour of chevron conditionally. I also used the `deactivate()` button method as otherwise it would still act as a button without performing any action.

@andrewtavis, could you please review these changes? 😊


andrewtavis:
Thank you, @SaurabhJamadagni! I’ll try to get to this tonight 😊😊

Issue #96, opened 2022-01-08
Convert JSON data to SQLite


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

There’s currently a bug with v1.1.0 where the Russian keyboard will not translate words. The keyboard switches to English as it should, and words can be typed, but then on pressing return the keyboard crashes without anything being inserted. 

### Device type

iPhone 7

### iOS version

 iOS 15.1

andrewtavis:
Is not reproducible on an iPhone 7 in Simulator, but then the version is 15.2.

japanese-goblinn:
I may take a look on this issue

andrewtavis:
This would be absolutely amazing! 🚀🚀 I really haven't had the bandwidth to dedicate time to this, but it's the main bug right now that dramatically affects the functionality. Let me know if you have ideas on this!

japanese-goblinn:
Yep, assign me on this and I'll let you know on my progress with this issue

andrewtavis:
Assigned! Thank you 😊

japanese-goblinn:
@andrewtavis I've looked intro this crash and will soon describe here what I've found

andrewtavis:
Thanks so much, @japanese-goblinn! Really looking forward to what you’ve figured out :)

andrewtavis:
@japanese-goblinn, note that I just checked and the new "lack of" behavior is reproducible on the Simulator, I just didn't check Russian after all these changes. Am looking into this as well now :)

japanese-goblinn:
Well, I managed to get this error while debugging, but not really sure how to reproduce it. I want to take a deeper look into this bug but unfortunately I don't have time for now.

This is what I found so far:

You sad that file is about `0.6 MB` when it's actually `15.1 MB`. Here's code how I measured it 
```swift
// func loadJSONToDict
let data = try Data(contentsOf: url)
let f = ByteCountFormatter()
f.allowedUnits = [.useMB]
f.countStyle = .memory  
print("[DEBUG]: file \(fileName) size \(f.string(fromByteCount: Int64(data.count)))")
```
[...]

PR #231, opened 2022-10-14
#96 add SwiftyJSON for data and refactor code

andrewtavis:
@SaurabhJamadagni, this is a PR that I think you should check in case I'm going in a wrong direction with how to solve #96. In the discussions there it looks like we need to do some kind of batch loading for the data, which will be something cool that you and I can look into 😊

What this does:
- Adds [SwiftyJSON](https://github.com/SwiftyJSON/SwiftyJSON) as a dependency, which in turn allows us to clean up a lot of the `?` and `!` optionals for data access
- Changes the data loading function to use it and removes the optional return for this function
    - The optional return was getting to be a massive headache, and we can guarantee the data will be there (see below)
- Makes dummy preposition JSON files for all languages that don't have them so that there's something to return

Lemme know what you think! From here the data calls are cleaned up a bit, which will make adding on further complexity for more advanced data loading a bit more reasonable :)


SaurabhJamadagni (approved):
Looks good to me @andrewtavis. The cleanup will definitely help. I think we need to move towards the idea of Language packs sooner than we expected to treat the root cause of large file sizes. Also, could you look into CoreData? I think that could help us reduce the sizes even further by having a data model hold the data instead of large data files. It hashes the data so we should automatically have smaller sizes. Let's talk about the next steps over a call sometime to figure out how to go ahead. SwiftyJSON seems like a step in the right direction to me.

andrewtavis:
> Hey @andrewtavis, I'll give this a look tomorrow morning. Sorry for the delay 😅

All good, @SaurabhJamadagni! I was out of town yesterday anyway 😊

> Also, could you look into CoreData?

Will look into it now! Thanks for the suggestion! 🚀

Issue #229, opened 2022-10-13
Regional keyboard variants


wkyoshida:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Opening this issue to ask about regional keyboard variants for the same language. 

As an example, some countries like France and Belgium use the [AZERTY](https://en.wikipedia.org/wiki/AZERTY) keyboard layout, the one implemented for the [Scribe French keyboard](https://github.com/scribe-org/Scribe-iOS/blob/524146390ae65a92026e5d7e3130558535f16236/Keyboards/LanguageKeyboards/French/FRInterfaceVariables.swift#L13-L16). French Canada, on the other hand, uses a [QWERTY](https://en.wikipedia.org/wiki/QWERTY#Canadian_Multilingual_Standard) layout. With that said, I was wondering if there were any thoughts on including regional variants as well. Using the example above, I could see someone in Canada likely wanting to use the QWERTY layout if their target language is French.

There can sometimes be minor differences even within layouts, e.g. the layouts for the Belgian and French AZERTY do. I do not think those differences would be as concerning, but at least the inclusion of distinct layouts, like AZERTY vs QWERTY, could be good perhaps. 

Curious to hear thoughts! My initial assumption is that this wouldn't be too difficult to implement for existing Scribe languages. I believe the core language-specific functionality could be reused and the main differences would be in the keys layout? If planned work comes out of this discussion, I can help. 

P.S. There can be other layout variants with other languages as well, particularly as more languages are added. I do not know all of them, but something to keep in mind.

andrewtavis:
Hi @wkyoshida! Thanks for the issue! I totally agree that regional variants would be really really important to include 😊 Hence the `priority` label. 

I originally blocked this as I was thinking that #16 might be necessary to create a menu item for selecting the layout, but then normally these are selected as separate keyboards :) Let me check to see if it would work to create a separate French keyboard and give it access to the French data files that are already there. If that can work, then as you said this should be fairly simple to implement! 🚀 If they can’t be accessed easily, then this might be tough as adding new keyboards does add a lot if size to the app. 

[...]

PR #236, opened 2022-10-17
#229 #96 adds french azerty and reduces JSON sizes

andrewtavis:
What's in here:
- All JSON files have had their indentation removed to reduce their filesizes (no need to check)
- Changelog updates
- Renames the French keyboard French AZERTY and adds French QWERTY
    - I made individual interface files that define the keys for each, which is then referenced within a common interface file that defines things like command button titles, alternate keys for long presses, etc (see `setFRKeyboardLayout`)
    - References to French are fixed throughout the project
- Fixes the `´` character on French keyboards
    - Also includes making sure it's sized correctly and doesn't shift with the letter keys


wkyoshida:
:raised_hands: will take a look at this!

andrewtavis:
Ah I forgot that you need to be a collaborator for me to request a review :) Let me know if that's of interest and we can definitely make that happen after a couple of contributions 😊 Thanks for taking a look!

Keyboards/LanguageKeyboards/French/FRInterfaceVariables.swift
```diff
@@ -6,106 +6,16 @@
...
-  uAlternateKeys = FrenchKeyboardConstants.uAlternateKeys
-  yAlternateKeys = FrenchKeyboardConstants.yAlternateKeys
-  cAlternateKeys = FrenchKeyboardConstants.cAlternateKeys
-  nAlternateKeys = FrenchKeyboardConstants.nAlternateKeys
-}
-
-/// Provides a French keyboard layout.
+/// Provides a French keyboard layout after checking for AZERTY or QWERTY.
 func setFRKeyboardLayout() {
-  getFRKeys()
```
wkyoshida, line 106: Oh I was thinking that `getFRKeys()` might be able to stay in this file, maybe no? Could it work to change `getFRKeys()` to take in a parameter for `FrenchKeyboardConstants` and then the `if` statements would just call it with either the AZERTY or QWERTY constants sent in? Perhaps you did try that already and there's just something I'm overlooking

Keyboards/KeyboardsBase/KeyboardKeys.swift
```diff
@@ -51,7 +51,11 @@ class KeyboardKey: UIButton {
       self.layer.setValue(true, forKey: "isSpecial")
```
wkyoshida, line 51: Just an observation on the changes in this file -

These are interesting, hmm.. :thinking: Makes sense that some keys should be ignored when it comes to upper/lower-casing. Just wondering if there could be a scenario where a key would have upper/lower-casing in one language but not in another. In that case, the logic here would depend on the target language. I'm thinking this would definitely be an edge-case, and I don't even know if it could happen :laughing: , but just a thought I had.

wkyoshida (commented):
I didn't go over the JSON files :eyes: but the French keyboard changes are looking good! Couple minor comments that I had :arrow_down:

Keyboards/KeyboardsBase/ScribeFunctionality/Conjugate.swift
```diff
@@ -236,7 +239,7 @@ func returnConjugation(keyPressed: UIButton, requestedForm: String) {
 /// Returns the conjugation state to its initial conjugation based on the keyboard language.
 func resetVerbConjugationState() {
   conjViewShiftButtonsState = .leftInactive
-  if controllerLanguage == "French" {
+  if ["French_AZERTY", "French_QWERTY"].contains(controllerLanguage) {
```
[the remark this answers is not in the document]
andrewtavis: Makes sense :) I think the easiest way of doing this right now that would allow for variants to be added without needing to change them is to just pattern match on the string in the following way:

```swift
controllerLanguage.prefix("French".count) == "French"
```

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Emojis for the following are chosen based on [gitmoji](https://gitmoji.dev/).
 
 # [Unreleased] Scribe-iOS 2.1.0
 
+### ⌨️ New Keyboards
+
+- Adds a QWERTY French keyboard for Canadian French.
```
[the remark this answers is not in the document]
andrewtavis: Great suggestion! 😊

Keyboards/LanguageKeyboards/French/French-AZERTY/FR-AZERTYInterfaceVariables.swift
```diff
@@ -0,0 +1,100 @@
+//
+//  FRInterfaceVariables.swift
```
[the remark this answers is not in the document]
andrewtavis: Thanks for the attention to detail!

Keyboards/LanguageKeyboards/French/FRInterfaceVariables.swift
andrewtavis: I made the decision to not include this with an eye on adding future ones that might have more distinct variations in the future, but thinking about it now I can't think of any that would be sooo different that we wouldn't be able to handle their variations in conditions at the top as we do with `setFRKeyboardLayout`. Thanks for bringing this up 😊

andrewtavis: But now I'm realizing that there actually is an issue here, as we'd need to make a `var` that can be assigned either of the constant enums, but then what happens is that this `var` would need to be instantiated with all the elements of the enums so that Swift/Xcode will know that they exist. I'm now getting errors that say that `FrenchKeyboardConstants.letterKeysPhone` doesn't exist. We should likely create a base enum that all the other interface variable ones are based on :) Would be a different issue though 😊

Keyboards/KeyboardsBase/KeyboardKeys.swift
andrewtavis: Thoughts of edge cases are much appreciated! There are all kinds of little things like that in Scribe 😊 Specifically `languagesWithCapitalizedNouns`, which as of now is `["German"]` 😄 We try to handle them as they come up :)

andrewtavis:
Thanks for the review, @wkyoshida! I'll merge this now and we can discuss a bit more in the issue 🚀

Keyboards/KeyboardsBase/ScribeFunctionality/Conjugate.swift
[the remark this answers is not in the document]
wkyoshida: Love it :+1: Yep, I agree! I think that'd be a pretty clean way of doing it if the variants all begin with the target language descriptor

Keyboards/LanguageKeyboards/French/FRInterfaceVariables.swift
wkyoshida: Ahh I see.. :disappointed:  I was concerned something like that might get in the way. 

Yep, I agree though. This can get refined later on in other issues, especially as work gets done on decoupling the language and the keyboard layout.

Keyboards/KeyboardsBase/KeyboardKeys.swift
wkyoshida: Oh.. the challenges of working with multiple human languages :laughing:

Keyboards/KeyboardsBase/ScribeFunctionality/Conjugate.swift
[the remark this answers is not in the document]
andrewtavis: Will have to, but I think it makes sense to do it that way anyway 😊

Keyboards/KeyboardsBase/KeyboardKeys.swift
andrewtavis: Welcome to Scribe! 🤣😊

Issue #181, opened 2022-07-23
Scribe-iOS Outreach


andrewtavis:
<a><img src="https://raw.githubusercontent.com/scribe-org/Organization/main/logo/ScribeAppLogo.png" width="25%" alt="Scribe Logo" align="right"></a>

## Sharing and promoting Scribe

This pinned issue is for discussing where and how best to promote Scribe-iOS. The purpose is to spread the word so that others can find Scribe as potential users, to provide feedback and to grow the contributors base. The issue text will be edited as the discussion below progresses :)

Everyone is welcome to share their inputs! Communities revolving around open-source, open-data, Wikidata, Wikimedia, Swift, iOS, iPadOS, language learning and all the languages Scribe supports should all be considered.

### Where Scribe is currently posted:

- [Wikimedia New Developers page](https://www.mediawiki.org/wiki/New_Developers)
- [Wikidata tools for lexicographical data](https://www.wikidata.org/wiki/Wikidata:Tools/Lexicographical_data)
- [open-source-ios-apps](https://github.com/dkhamsing/open-source-ios-apps)
- [up-for-grabs.net](https://github.com/up-for-grabs/up-for-grabs.net) [(pending)](https://github.com/up-for-grabs/up-for-grabs.net/pull/3271)

### Organizations that have featured Scribe:

- [Wikimedia](https://www.wikimedia.org/)
- [Wikimedia Deutschland](https://www.wikimedia.de/)
- [CocoaHeads Berlin](https://www.meetup.com/de-DE/cocoaheads-berlin/)

### General

- [Twitter](https://twitter.com/scribe_org) - suggested retweets and posts welcome!
- [Hacker News](https://news.ycombinator.com/)
- [Product Hunt](https://www.producthunt.com/)

### Tech Subreddits

|                                                        |                                                  |                                            |
| :----------------------------------------------------- | :----------------------------------------------- | :----------------------------------------- |
| [r/programming](https://www.reddit.com/r/programming/) | [r/software](https://www.reddit.com/r/software/)                                            | [r/apple](https://www.reddit.com/r/apple/)           |
| [r/ios](https://www.reddit.com/r/ios/)                 |                                                  [r/swift](https://www.reddit.com/r/swift/) | [r/opensource](https://www.reddit.com/r/opensource/) |

### Language Learning Subreddits

|                                                                  |                                                                |                                                      |
| :--------------------------------------------------------------- | :------------------------------------------------------------- | :--------------------------------------------------- |
[...]

PR #239, opened 2022-10-21
#181 add Twitter and Reddit buttons to repo readme

andrewtavis:
Here's my attempt at making them work, @SaurabhJamadagni 😊 I chose to just add them to the end of the top shields as we're just trying to prompt people when they land there to consider using them or sharing in their own way. I don't think that we need to put them on their own line :)

The twitter button has text with `#languagelearning` and the reddit button just has the URL and a general title, which I thought made sense as we're not sure where they'll post it. Your input is more than welcome!


SaurabhJamadagni (approved):
Works great @andrewtavis! I pushed it to my fork and the buttons look great on the README. Perfect location and the post it creates when clicked looks good too!

andrewtavis:
Thanks, @SaurabhJamadagni! Will bring this in now then :)

Hope you're having a nice weekend! 😊

Issue #208, opened 2022-09-12
Default autosuggestions for pronouns


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Once work is finished on #194, another feature that can be added is to create baseline autosuggestions for when a user enters in a pronoun. This would require making sure that all the needed conjugations are present on Wikidata, but generally what we'd be looking for is always suggesting the conjugations of "have," "be" and "can" when a user types a pronoun. This has been found to be a pattern in system keyboards, and making sure that there's some consistency to these very common words makes sense (what words follow pronouns in each language should be checked though).

We'd need to make a variable `pronouns` in [KeyboardsBase/InterfaceVariables.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/InterfaceVariables.swift), and this array would be overwritten in the `set{languageCode}KeyboardLayout` functions with pronouns for each language within their respective `InterfaceVariables.swift` files. We'd then be able to check if one of these words has been typed, and then return the conjugations within `getAutosuggestions` in [KeyboardViewController.swift](https://github.com/scribe-org/Scribe-iOS/blob/main/Keyboards/KeyboardsBase/KeyboardViewController.swift).

- Currently blocked by #194

### Contribution

I'd be happy to work on this or direct someone in a contribution once #194 is finished.

SaurabhJamadagni:
Hey @andrewtavis, I would like to work on this issue. Seems like the extension of what I have already been doing haha. Just to clarify, as we did for placeholder text, we'd have to create a pronouns array for each language. Then if the word typed by the user is in the pronouns array, the correct conjugation of [have, be, can] should be suggested. I am unable to understand how the right conjugation would be suggested. Will it require additional data or will we need to use the `autosuggestions.json` file for it? Could you help me out with this? I'll get to it immediately.

andrewtavis:
This seems like a reasonable one for me too :) 

[...]

PR #240, opened 2022-10-22
Including a pronouns to tense dictionary for French

SaurabhJamadagni:
- PR for sub task of issue #208 
- Added a dictionary in french interface variables.

@andrewtavis could you please review these changes. Also, maybe confirm that the dictionary is correct. I created it using as much I could understand about French pronouns from an article 😅

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -271,7 +271,7 @@ class KeyboardViewController: UIInputViewController {
     completionWords = [String]()
     var i = 0
     while i < 3 {
-      let suggestion = verbs[verbsAfterPronounsArray[i]][pronounAutosuggestionTenses[prefix.lowercased()]!].string ?? ""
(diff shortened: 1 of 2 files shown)


andrewtavis (changes requested):
All looks good, @SaurabhJamadagni! 😊 Changes requested, as don't we need to add a check into `getPronounAutosuggestions` to see if it is a verb first, and if not just return the word? `ne` - French for `not` - won't be conjugated, so we just want this word returned :)

andrewtavis:
The above commits also fixed the super minor issues with the French files :) Thanks for working with the language files like this! 🚀

SaurabhJamadagni:
@andrewtavis ce4b7b6 should fix it. As the `verbs` JSON returns an optional, I think it should do it. If no value is found, it by defualt returns the verb itself. Will this do?

andrewtavis (approved):
Fixes are perfect, @SaurabhJamadagni! Thanks for the work on this :) :) We can do a similar bit for Spanish for the words listed in the issue, and then we're done with this 🚀

PR #244, opened 2022-10-25
Added functionality to have the keyboard first display the language on the space bar before switching to the language's word for "space" (note, currently doesn't animate from one to the other like the system keyboard, kinda just pops from one to the other)

helloITdepartment:


andrewtavis:
Checking all this now, @helloITdepartment :) Likely best if you do a branch for each change going forward, but also totally fine for these first contributions 😊 It just looks like the changes here are also propagated into #245 and #247, where branches would keep them separate :)

andrewtavis:
@helloITdepartment, I'm actually getting an error when I try to run the code at this point. Is saying that it can't find `setBtn` and other btn functions in the main view controller 🤔 Maybe you could check if it's working on your end as well? A bit weird as it was fine when we looked at it. Also, there's spot in there where `languageTextForSpaceBar` has a lower case `b` for `Bar` at the end. This might be a bit weird as you're just working on one branch for this and the other two PRs, so if you make a change here you'd need to update those as well 🤔 

I'd say let's first see if you get an error right now, and then if so we can brainstorm on it. I looked for a bit and it didn't jump out to me, so I figured I'd check in :)

Hope you're having a nice week!

andrewtavis:
NVM, @helloITdepartment 😊 I figured it out :) Just needed to see the exact changes. There's also a bug where typing anything also shows the keyboard language, but I fixed that by changing `hasShownKeyboardLanguage` to `showKeyboardLanguage` and only setting it to true at the part in load keys where things that are assigned on the initial run are done. Let me make the edits on this now :) :)

andrewtavis (approved):
Sorry for the poor code review conventions here. I need to learn how to checkout someone else's code and commit directly to it. With that being said, This is the code I have running on my machine, and it's working great! 😊 Thanks for the hard work on this. Ended up being a bit more annoying than we expected, but the result is great! 🚀

PR #247, opened 2022-10-27
Fix: allow for autocomplete after certain punctuation -- ", /, #, (

cdmill:
Added `currentPrefix.hasPrefix()` checks for `"`, `/`, `#`, and `(`. If the word begins with any of those punctuation, then `replaceOccurrences` will remove it in order to get the autocompletions. If we didn't wan't to call `currentPrefix.hasPrefix()` multiple times in the if statement, another workaround could be to store the initial character of the word as a constant and then check if the constant is `"`, `/`, `#`, or `(`. If you would rather do it that way or another way let me know and I can change it.


andrewtavis (approved):
Works like a charm! Thanks, @cdmill! 😊

Issue #250, opened 2022-11-06
Highlight autocompletion if it is the word typed


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue a a simple one to reflect system behavior where when the keyboard suggests the exact word the user has typed as a completion it's usually enclosed in quotes. What Scribe could do for this is lightly change the background of the autocompletion if the word is the given word. This is also done in some other keyboard apps, and I think it's more appealing than the word in quotes. 

The background of the completion key would thus need to be conditionally changed and reset as the user continues typing :)

### Contribution

Happy to support someone who wants to try this issue 😊

techsandu:
Hey @andrewtavis can you please assign this in my name? I will look into this and come back.

andrewtavis:
Sure thing, @techsandu! Thanks for reaching out! 

Would you like me to give you a detailed breakdown of where to look in the codes? Happy to send something along to contextualize this for you 😊

andrewtavis:
In case it would be helpful, the assignment of the word is within this function:
https://github.com/scribe-org/Scribe-iOS/blob/fb268f43383a0215dd74ddc8648ea35b493503f1/Keyboards/KeyboardsBase/KeyboardViewController.swift#L360

So basically we need to do a comparison of the word that's in the proxy with `completionWords[0]`, as naturally the first suggestion would be the one we want a slightly different coloration for given alphabetization. If memory serves me, `currentPrefix` also saves the value being typed in any given moment, so that would be what we're comparing `completionWords[0]` to, and in case they are the same we'd be changing the `color` argument of the function `setBtn` in `conditionallySetAutoActionBtns` linked to above 😊

As far as a color, I'm open to suggestions, but maybe something like `keyColor` with an alpha would be in order? Would be very happy to answer any other questions you have about this!

andrewtavis:
@techsandu, checking in with you on this :) Let me know if you need any assistance 😊

cameronslee23:
Hi @andrewtavis, Could you assign this issue to me? This is my first contribution to an open source project and I have a pull request ready to submit. I'm still a little bit confused about the process of contributing so any guidance would be much appreciated!

[...]

PR #253, opened 2022-11-27
Added support for highlight autocompletion

cameronslee23:
### Contributor checklist

- [x] I have updated the [CHANGELOG](https://github.com/scribe-org/Scribe-iOS/blob/main/CHANGELOG.md) with a description of my changes for the upcoming release 

---

### Description

This pull request proposes to add support for highlighting the "AutoAction1" button given that the word being entered is the completion word at hand. The color of the button is changed to "keyColor" and given an alpha component to distinguish itself.

I tested that this change works by running Scribe in an emulator and typing the word listed in the "AutoAction1" button until they were a match. I then confirmed that the button was highlighted only when the word I had typed was the same as the word in the "AutoAction1" 

### Related issue

#250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Emojis for the following are chosen based on [gitmoji](https://gitmoji.dev/).
 ### 🗃️ Data Added
 
 ### ✨ New Features
+- Added highlight for autocompletion if it is the word typed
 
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Well done, @cameronslee23! Is exactly what I was thinking :) Thanks so much 🙏

Issue #103, opened 2022-01-20
Make return button read search in browsers


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Currently Scribe keyboards don't change the return button to something like `go` or `search` when in a browser. It's possible that not many users would find Scribe helpful when searching the web, but as of now it functions well in that commands can be called, and after confirming the choice the search is updated. This is something that is standard in system keyboards, and should be included in Scribe if possible.

Initial inspections of this indicate that [textFieldShouldReturn](https://developer.apple.com/documentation/uikit/uitextfielddelegate/1619603-textfieldshouldreturn) might be a trigger that could be used in these cases. 

Possibly helpful SO questions are [here](https://stackoverflow.com/questions/50444197/how-to-add-target-to-keyboard-done-go-search-buttons), [here](https://stackoverflow.com/questions/41059449/uisearchbar-detect-when-user-stop-type-in-swift), [here](https://stackoverflow.com/questions/30415284/change-uisearchbar-keyboard-search-button-title-in-swift) and [here](https://stackoverflow.com/questions/31886720/change-return-button-function-to-done-in-swift-in-uitextview).

andrewtavis:
Whether the button should change color in these cases should also be discussed. If it is, changing it to the blue that the system keyboard uses for light and dark mode would be best.

andrewtavis:
It appears that checking for if an input is `UIReturnKeyType.Done` could be the needed conditional to change the return key over.

gitpushOmnik:
Hey @andrewtavis, I would love to work on this issue as well. Can you assign it to me please?

gitpushOmnik:
Yes please. That would be helpful to get me started. I am researching on it myself as well

andrewtavis:
Here are some things that came up recently, @gitpushOmnik:

- As stated [above](https://github.com/scribe-org/Scribe-iOS/issues/103#issuecomment-1181680775) we can check if the return key type is done, but then I'm not sure if this is being transmitted or not 🤔
  - I found this in [this SO question](https://stackoverflow.com/questions/30415284/change-uisearchbar-keyboard-search-button-title-in-swift)
[...]

PR #267, opened 2023-03-23
Read Search for Return Button

gitpushOmnik:
### Contributor checklist

- [x] I have updated the [CHANGELOG](https://github.com/scribe-org/Scribe-iOS/blob/main/CHANGELOG.md) with a description of my changes for the upcoming release (if necessary) 

---

### Description

- This pull request contains commits for changing the title of the return button to "search" whenever the keyboard is being used by the user in the browser. 
- The return button title automatically changes to its default symbol when the user is not using a web broswer.
- The code has been tested on Safari

### Related issue

#ISSUE_NUMBER
This pull request targets #103


andrewtavis:
Hey @gitpushOmnik! Thanks so much for the PR! Let me just check one thing with you prior to jumping in :) As I understand from your comments in #103, `webSearch` being the basis for this means that this should only be triggered by it being in the search bar itself, right? The functionality is not that it switches to `search` or `go` for all cases in a web browser like when filling out forms and such?

Will say it here once and say it in the issue again, really appreciate all the research you did here! It's a simple thing, but it is expected behavior for keyboards, so this'd be great to figure out! 😊

andrewtavis (approved):
Looking at it now, @gitpushOmnik, and it seems good to me :) We can't figure out if something on a webpage is a search field, but we can figure out if it's a main search field for a browser. I'm kind of considering making it less "intrusive" now that it's impossible to do it across searches. What might make more sense is to just maintain the `arrow.turn.down.left` and simply change the color of the button to `linkBlue` 🤔 That way the functionality from the system is in there, but then someone isn't confused when the button doesn't change its text. Also makes it easier to not have to format the size of the text based on how large the languages word is 😉

Thanks for this! I'll bring it in and play around a bit and then write to you in the issue :)

gitpushOmnik:
Thank you so much @andrewtavis

Issue #51, opened 2021-11-24
Add emojis to autosuggestions


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to discuss and eventually implement the addition of emojis to autosuggestions. This would likely necessitate the addition of emojis as lexemes to [Wikidata](https://www.wikidata.org/) along with the words that represent them in each language. Typing any of these words would then replace the last autosuggestions with emoji options. Whether Scribe should have a dedicated emoji button and an emoji search could be covered in another issue, but the Scribe UI could lend itself to making this a more seamless process.

- This feature is blocked by #188 and #194 as Scribe first needs autosuggestion functionality

andrewtavis:
A way that Scribe could improve on the performance of the system keyboard would be to offer suggestions even after the text has been broken by punctuation. So for example you could still suggest emojis after `"Thank you!!!!"` with `🙏 ` being the logical suggestion.

andrewtavis:
Creating arrays of words and their emoji equivalents might be something that could also be be done in Python within [Scribe-Data](https://github.com/scribe-org/Scribe-Data). From there autosuggest could check if the word is in the emoji array, and then two emojis from the associated list could be suggested. At this time the emoji suggestions would definitely be two instead of three as sometimes can be the case, as the Scribe key restricts the available space too much.

This issue could be unblocked and an issue in Scribe-Data could be made after #194 is finished :)

andrewtavis:
Adding some easter eggs into this feature could also be a nice touch. These might be best as just autosuggestions and not corrections. Some ideas are:
- Wikidata: 💚 and 📁
- Wikipedia: ❤️ and 📚
- Scribe: 💙 and ✏️

andrewtavis:
@SaurabhJamadagni, FYI we have the data for emoji autosuggestions now via 1cc1ea5 🥳🙌

andrewtavis:
[...]

PR #276, opened 2023-04-01
Adding Emoji autosuggest

SaurabhJamadagni:
### Contributor checklist

- [x] I have updated the [CHANGELOG](https://github.com/scribe-org/Scribe-iOS/blob/main/CHANGELOG.md) with a description of my changes for the upcoming release (if necessary) 

---

### Description

- Added two extra buttons to command bar that are used for emojis. These are deactivated when autosuggest is active.
- A simple loop appends the top two emojis for a given prefix. These emojis are then assigned as title to the emoji suggestion buttons.
- Reused as many functions as possible to avoid writing redundant code.
- Tested on iPhone 14 Pro, iPhone SE (3rd Gen) and iPad Pro (11")

@andrewtavis could you take a look at these changes? Thanks for your help along the way :)

### Related issue

- PR for issue #51

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ Emojis for the following are chosen based on [gitmoji](https://gitmoji.dev/).
 ### ✨ New Features
 
 <!-- Scribe data is now loaded into SQLite database tables to make data reference less memory intensive and mitigate crashes. -->
-<!-- Emoji autosuggestions are now available as the user types. -->
 
+- Emoji autosuggestions are now available as the user types. 
 - Added action to command bar information icon.
 - Added highlight for autocompletion if it is the word typed.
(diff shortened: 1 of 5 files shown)


andrewtavis (approved):
Works like a charm, @SaurabhJamadagni! 👍🥳🚀 Only change that I made was resetting the emojis each time, as the way it was set up the emojis would be appended to a list that was never rewritten, so the first emoji suggestions someone got would be the only ones they'd see 🙃

andrewtavis:
Will take this in and do some minor changes to it all for styling and such :)

SaurabhJamadagni:
> Only change that I made was resetting the emojis each time,

Ahhh, I think I only reset it when the auto action keys were pressed. Should have also reset it when spacebar was pressed to begin a new word. Thanks for the fix @andrewtavis! 😄

Issue #308, opened 2023-05-21
Add period and comma to letter keys on menu selection


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to implement the change to the keyboard that would be triggered by the menu item in #196. If the user chooses to include a period and comma in the letter keys via the menu, then a period would appear to the left of the spacebar and a comma to the right of the spacebar for letter keys, number keys and special keys.

- This issue is blocked by #196 

### Contribution

This issue will be done as a part of Google Summer of Code 2023 :) Happy to support where needed!

SaurabhJamadagni:
Hey @andrewtavis! I just wanted to update you on the progress of this issue. I was able to figure out how to check if the user has the extensions installed. I also figured out establishing common data stores so different targets can save their preferences.

I haven't pushed the changes yet as it's a bit buggy right now and I'm trying to identify that problem. To be specific, currently it won't reload the list when a user adds or removes a language from Settings. Trying to figure out where would be a good place to refresh. 

I am sorry on the delay. But I'm a bit occupied with last minute shopping and packing for the move. I'll share the changes soon. Just wanted to inform about the delay 😅

andrewtavis:
No stress, @SaurabhJamadagni! Appreciate the efforts and am looking forward to the changes! 🚀 I definitely understand that it’s tough to find a spot to refresh sometimes :) 

Good luck with the move! Hope that it all goes well 🚚😊

andrewtavis:
Resolved via #340 :) :) Thanks @SaurabhJamadagni!

Issue #196, opened 2022-08-23
Menu item to include period and comma in letter keys


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something that Android users who switch to iOS usually want is the ability to have a period and a comma on the main keyboard so that they don’t need to press ‘123’ to reach them. An option could be included in Scribe’s menu to allow these keys to be included, with each being added by a separate switch. 

This would be a boolean option where the characters would be added to the array of letter keyboard characters for the keyboards that come from the interface variable files.  

- This issue is currently blocked by #16

### Contribution

I’d be happy to work on this when it’s unblocked :)

japanese-goblinn:
@andrewtavis is this an active issue that I can work on? What is `#men` in `This issue is currently blocked by #men`? Maybe you can suggest more relevant issues that I can work on?

andrewtavis:
> What is `#men` in `This issue is currently blocked by #men`?

What that is is "This issue is currently blocked by #what's that issue about the menu lemme type men and select the auto complete but then it doesn't give me the auto complete so I just get 'men' and don't realize it" 😄

We need to have a menu before this can be worked on, hence #16 needs to be done (fixed). This is a lot of work, so let's finish the current version and then we can move on to that and this issue as well. There is a lot of interest for this, so let's definitely keep it in mind :)

I'll write here with some others that you could consider 😊

andrewtavis:
Regarding issues that might be of good to work on, you now understand a bit of what's going on under the hood for switching conjugations, so https://github.com/scribe-org/Scribe-iOS/issues/211 might be a good fit? That's greying out the left or right button if we're at one of the points where the break would occur in the new enums and disabling the button :) Would definitely be more involved, but I'd be happy to write some directions in there :)

andrewtavis:
Note on this issue: as iPads have periods and commas on the letter keys, this option shouldn’t show up on iPads :)

andrewtavis:
Resolved in the referenced #340 🙃 Thanks @SaurabhJamadagni! 🙌

PR #340, opened 2023-07-16
Adding a Comma and Period option to the settings menu

SaurabhJamadagni:
### Contributor checklist

---

### Description

Hey @andrewtavis, here's one more giant PR again unfortunately 🙈 

The PR adds:
- Lists the installed keyboards
- Displays menu option to toggle between comma and period for keyboards
- Translates the preferences to the keyboard view controller on launching a particular language keyboard.

Regarding the problem I was facing about refreshing the list, it has been worked out in a very hacky manner for the moment. I will explain the problem better in the meeting tomorrow. The gist is that it works when we open the app on some other menu tab and then navigate to the `Settings` tab. This allows me to call the `viewWillAppear` method which takes care of the refresh. The refresh uses the Notification system. Publishes a notification whenever the value changes which is then handled by a function that is set to execute. This function calls the refresh on the child table views. 

I also tried to add a notification observer to only the dynamic data sections so that not all child tables are reloaded. This is done using an additional property `hasDynamicData` which accepts an enum value. Chose an enum since if we add more sections further that have dynamic data, handling it using a switch statement should be easier.

I plan on working on this in the background. But this issue was holding back from moving on to the next issues so I decided to push this unoptimised yet best working state. Do let me know if it is okay for now.

### Related issue

- #196 
- #308

### DIFF
--- a/French.entitlements
+++ b/French.entitlements
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.scribe.userDefaultsContainer</string>
+	</array>
+</dict>
+</plist>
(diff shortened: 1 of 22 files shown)


andrewtavis (approved):
Changes look good to me, @SaurabhJamadagni 😊 Obviously the hacky behavior of jumping to `Installation` from `Settings` on reload is not ideal, but let's just be sure to have this fixed in the coming weeks when we get ready to release `v3.0.0`. I can also try to look into this when I re-implement the German app localization :)

SaurabhJamadagni:
>  let's just be sure to have this fixed in the coming weeks

@andrewtavis, could you open up an issue for it? I think it would be good to document the changes that follow in case we come across a bug similar to this down the line. Also, should serve as a reminder for us as well. Or I'll create the issue tomorrow if you are occupied. No worries 😄

andrewtavis:
I’ll make the issue now, @SaurabhJamadagni  :)

andrewtavis:
See #344, @SaurabhJamadagni 😊

Issue #337, opened 2023-07-03
Swipe between parent views within menu


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would allow a user to swipe between parent views in the Scribe menu. The way it would work is that the user would be able to swipe back when in a child screen of the menu like for keyboard settings, but if they're on the main `Installation`, `Settings` or `About` screens then swiping would allow them to switch to one of the other tabs.

### Contribution

Happy to help with this if it's needed! This issue is assigned to @SaurabhJamadagni as a part of GSoC 2023, but if someone else wants to work on it then he could also review the contribution :)

nyfz18:
Hi!

andrewtavis:
Hi @nyfz18 👋 I screwed up and said #337 when I should have said the other one we talked about - #33 :) This one is the one that would require you to have a simulator 🙃 We can check out #33 for now and then figure out the next thing to do after that!

nyfz18:
sure! I'll take a look at it

SaurabhJamadagni:
Hey @andrewtavis, for this issue I was wondering if we can go with a package that will allow for this functionality. Here's the [link](https://github.com/marcosgriselli/SwipeableTabBarController/tree/master) to it. The animation based approach I talked about last week just doesn't feel natural. Any other approach I have seen will need a lot of reengineering. This can be a difficult change since a Tab Bar is close to the root view controller on top of which we have the rest of the screens in the Storyboard file. Can we give this package a try? Maybe it could be useful.

andrewtavis:
Looks good to me, @SaurabhJamadagni! Feel free to implement using `SwipeableTabBarController` :)

SaurabhJamadagni:
Hey @andrewtavis just to confirm, what have you used to install the swiftyJSON and GRDB packages? Was it swift package manager?

andrewtavis:
Hey @SaurabhJamadagni 👋 Yes, I used Swift Package Manager :)

andrewtavis:
[...]

PR #349, opened 2023-08-10
Swiping between tabs using SwipeableTabBarController

SaurabhJamadagni:
### Contributor checklist

---

### Description

This PR adds the ability to swipe between menu tabs. It uses the [SwipeableTabBarController](https://github.com/marcosgriselli/SwipeableTabBarController) package. 

@andrewtavis could you test it out on a physical device if possible? Seems to work great on the simulator. I would have but then for me to run it on my device it will have to go and provision certificates for all languages and that's causing some problem.

### Related issue

- #337


andrewtavis (approved):
Looks great to me, @SaurabhJamadagni :) I don't have testing setup on a physical device, but let's just role with this and we can react to it if it's not working locally. If this is concerning for you (which I totally agree it's not best practice), then I can set something up before release. No stress 😊

Thanks for this!

Issue #366, opened 2023-10-05
Multiple Delete Keys on IPad


stilianb:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

In the current main branch, all keyboards (not including Hebrew, Danish, and Norwegian) have multiple delete keys within their respective interfaces on IPad.

Here's an example of the issue:
<img width="492" alt="deletekeybug" src="https://github.com/scribe-org/Scribe-iOS/assets/49213751/c0e8e45e-fdaa-4b30-b618-652454c0c8d0">

`letterKeys[0].append("delete")` within each languages interface seems to be appending another delete key for IPad layouts.

The earliest instance I could find of this change being made was in #33.

Removing this line in each interface seems to solve the issue.

### Device type

iPad

### Versions

iOS 17, Scribe 2.3.0

andrewtavis:
Hey @stilianb! Thanks so much for pointing this out 😊 Would you have interest in working on this?

stilianb:
Yes! I'd be happy to.

andrewtavis:
Fantastic, @stilianb! Assigning you now :)

andrewtavis:
Closed by #367! Thanks for opening the issue and sending along a fix, @stilianb! As I said in the PR, I'll finalize these conditions soon and then put this addition of the delete key back in a context that it's needed.

PR #367, opened 2023-10-10
Removing multiple delete keys on IPad

stilianb:
### Contributor checklist

---

### Description

- As shown in the related issue, the current version of Scribe-iOS appends an extra delete key within each languages interface, specifically for IPad. The solution proposed in this PR removes said append in each interface.

- Testing for this change was done on an emulator for IPad and IPad Mini using Xcode.

### Related issue

- #366

### DIFF
--- a/Keyboards/LanguageKeyboards/English/ENInterfaceVariables.swift
+++ b/Keyboards/LanguageKeyboards/English/ENInterfaceVariables.swift
@@ -104,7 +104,6 @@ func getENKeys() {
 
     // If the iPad is too small to have a numbers row.
     letterKeys.removeFirst(1)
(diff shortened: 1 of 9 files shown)


andrewtavis:
Thank you for this, @stilianb! Will try to get to it in the coming days :)

andrewtavis (approved):
Thank you @stilianb :) These lines here are going to be needed again soon, but then they're meant to be within a conditional that would include them in cases where we eventually don't have a numbers key row on the letter keys and then do want to move the delete key down to what would then be the first row.

I'm actually working on these conditions now, so I'll bring this in and then fix it :)

Issue #372, opened 2023-10-21
Remove `Disable accent characters` for menus that it's not needed on


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

 In #339 we added an option to disable the accent characters of certain keyboards. Something that we need to do now is have this option show up in the settings only for those languages that it makes a change for: specifically German, Swedish and Spanish. 

Would be great if someone wants to help with this! Happy to support or get to it soon 😊

### Device type

iPhone and iPad

### Versions

Current development branch

henrikth93:
I am interested in this one

andrewtavis:
Thanks @henrikth93! Assigning you now :)

andrewtavis:
It took just a bit for the hack to really grow on me, @henrikth93 :) d2e4c67 sends along formatting for the code, but aside from that all's great. Makes sense to me. We want an index, we need a default in case it's not there, and we assign -1 so we don't need to deal with converting the type. Thanks for the work here and the great idea! 😊

PR #375, opened 2023-10-25
Remove Disable accent characters for menus that it's not needed on #372

henrikth93:
### Contributor checklist

---

### Related issue

- #ISSUE_NUMBER

### DIFF
--- a/Scribe/Components/ParentTableViewCell/ParentTableViewCell.swift
+++ b/Scribe/Components/ParentTableViewCell/ParentTableViewCell.swift
@@ -133,6 +133,11 @@ extension ParentTableViewCell: UITableViewDelegate {
       case .appLang: break
       case .specificLang:
         if let viewController = parentViewController?.storyboard?.instantiateViewController(identifier: "TableViewTemplateViewController") as? TableViewTemplateViewController {
+          let accentKeyLanguages: [String] = ["Swedish", "German", "Spanish"]; //Languages where we can disable accent keys
+          if (accentKeyLanguages.firstIndex(of: section.sectionTitle) == nil)
+          {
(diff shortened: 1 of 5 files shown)


Scribe/Components/ParentTableViewCell/ParentTableViewCell.swift
```diff
@@ -133,6 +133,11 @@ extension ParentTableViewCell: UITableViewDelegate {
       case .appLang: break
       case .specificLang:
         if let viewController = parentViewController?.storyboard?.instantiateViewController(identifier: "TableViewTemplateViewController") as? TableViewTemplateViewController {
+          let accentKeyLanguages: [String] = ["Swedish", "German", "Spanish"]; //Languages where we can disable accent keys
+          if (accentKeyLanguages.firstIndex(of: section.sectionTitle) == nil)
+          {
+            SettingsTableData.languageSettingsData[0].section.remove(at: 1) //If there are no accent keys we remove the option.
```
andrewtavis, line 139: This is really great, @henrikth93! Could we add one line right before this that checks what the the index is for `Disable accent characters` and then we can drop that index specifically? As of now if something else goes in at index 1 this will break its functionality :)

Aside from this, 100% what we're looking for! 😊

andrewtavis: Thanks @henrikth93! Will merge it in soon after 🤠

andrewtavis (approved):
I see what you mean by hacky, @henrikth93 🙃 Looks good though! I'll bring it in and see if I can do something to get rid of the need to do the `-1` option :) Will update in the issue if I can figure something out!

PR #381, opened 2023-11-04
Refactor: Extract DB access logic from functionalities and views

flumaves:
### Contributor checklist

---

### Description

- Rename loadData.swift to LanguageDBManager.swift
- About LanguageDBManager, the DB is accessed from multiple places externally, so designing it as a singleton is a suitable method.

Notes about LanguageDBManager
- Direct operations on the database are not allowed and can only be completed by the manager.
- Extract database access in views and functions into separate methods as an extension to LanguageDBManager.
- SQL query statements should be written in the extension method. Adding new methods should also be written in the extension and avoid duplicate methods.


andrewtavis (approved):
Apologies that this review took so long and that the merge is as messy as it's going to have to be, @flumaves. There's been lots going on that I've needed to direct my attention to. It was not my intention to have this take as long as did.

There are some issues with some features breaking with the new changes. Didn't have the capacity to deal with them till now, but now the merge conflicts have grown. I'll bring in the new db file and then will go through the commits here to rework the conversion of the SQLite calls resembling your approaches.

Appreciate the contribution! Hope that all's well :)

Issue #25, opened 2021-11-23
Fix transition from landscape to portrait


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

This issue is to fix the transition from portrait to landscape so that height is re-rendered. As of now when a user switches orientations the height of the keyboard as defined by `keyboardHeight` is not reset, which cases the keys to be too big or small depending on which transition is being made. Basically the `keyboardHeight` of the initial view is maintained regardless of the orientation, and the only way to reset it as of now is to reselect the keyboard or to close and open it again.

### Device type

iPhone and iPad

### iOS version

I assume this is happening for all versions.

andrewtavis:
The issue now just deals with a poor transition from landscape to portrait where the keys are smaller than they should be.

andrewtavis:
Closed via #388 😊 Thanks @henrikth93! Love getting these early issues taken care of 🚀

PR #388, opened 2023-12-11
Update KeyboardViewController.swift

henrikth93:
I added timers that makes sure the width and height settings are correct when rotating.

### Contributor checklist

---

### Related issue

- #25


andrewtavis:
Thanks @henrikth93! I'll take a look soon :)

andrewtavis:
Some thoughts on this, @henrikth93:

- The transition is great!
- Sometimes the keys are not exactly sized the way they should be, but it's a major improvement
- Looks like it might be good to call the annotation function one more time after the transition?
  - We can discuss in the sync, but for me with the German keyboard the annotation bar under words shifts afterwards, but after I start typing it looks ok again (see screenshots)

Simulator Screenshot - iPhone 15 Pro - 2024-01-13 at 14 17 44

Simulator Screenshot - iPhone 15 Pro - 2024-01-13 at 14 18 04

After fixing that we should be good to go! 🚀

andrewtavis (approved):
Merging this as the comments on annotation problems were found to be unrelated to the changes here :) Thanks for the fix, @henrikth93! So great to have this working better at this point 😊

Issue #362, opened 2023-09-18
Size keys on expanded iPad keyboards


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is the second half of #352 where the expanded keyboard layouts for iPads are assigned and are able to be switched to with the `123?!` key. Within this issue we'll go through and create conditionals to style the keyboards based on the new keys such as the tab key and others :)

Note that we'll be going for a generic layout for all keyboards. In the system keyboard sometimes there are differences like the return key being L shaped and other elements, but this will not be done at this time. Goal for this is to include this in the 3.0.0 release 🚀

### Contribution

Happy to support someone with interest in working on this!

andrewtavis:
CC @henrikth93 who's working on #352 :)

andrewtavis:
Thank you, @henrikth93! Assigning now 🚀🚀

andrewtavis:
Hey there @henrikth93 👋 Here are the layouts that we're looking for in one place that I'm bringing over from #33:

## Danish (#133)

![Simulator Screenshot - iPad Pro (12 9-inch) (4th generation) - 2023-07-12 at 21 57 51](https://github.com/scribe-org/Scribe-iOS/assets/24387426/fe160c02-0a4c-4a4b-ac56-dad97894b8cb)

![Simulator Screenshot - iPad Pro (12 9-inch) (4th generation) - 2023-07-12 at 22 15 28](https://github.com/scribe-org/Scribe-iOS/assets/24387426/c6c26bcc-cd12-49ee-a9f6-bbefb4e24f7f)

## English (#7)

![Simulator Screenshot - iPad Pro (12 9-inch) (4th generation) - 2023-07-12 at 22 00 15](https://github.com/scribe-org/Scribe-iOS/assets/24387426/6dfe5775-98af-4339-be3d-6fc62b0b278e)

![Simulator Screenshot - iPad Pro (12 9-inch) (4th generation) - 2023-07-12 at 22 16 16](https://github.com/scribe-org/Scribe-iOS/assets/24387426/d4e7ac2d-b440-45f6-9541-2ecaa7d0f2af)

## French

### Canada

[...]

PR #397, opened 2024-01-28
#362 Changed the expanded keyboard for more languages

henrikth93:
### Contributor checklist

---

### Related issue

- #362


Keyboards/KeyboardsBase/KeyboardKeys.swift
```diff
@@ -254,31 +254,60 @@ class KeyboardKey: UIButton {
...
+        scalarSpecialKeysWidth = (disableAccentCharacters && keyboardState != .symbols) ? 2.2 : 1.0
+        scalarReturnKeyWidth = (disableAccentCharacters && keyboardState != .symbols) ? 1.5 : 1.0
+        scalarCapsLockKeyWidth = 1.3
+        scalarShiftKeyWidth = 1.8
+      default:
+        scalarSpecialKeysWidth = (disableAccentCharacters && keyboardState != .symbols) ? 2.2 : 1.0
+        scalarReturnKeyWidth = (disableAccentCharacters && keyboardState != .symbols) ? 2.2 : 1.0
+      }
+      //scalarSpecialKeysWidth = (disableAccentCharacters && keyboardState != .symbols) ? 2.2 : 1.0
+      //scalarReturnKeyWidth = (disableAccentCharacters && keyboardState != .symbols) ? 2.2 : 1.0
```
wkyoshida, line 284: nit - Would it make sense to already check `(disableAccentCharacters && keyboardState != .symbols)` just once before and then simply use the boolean result in the `switch` statement? The boolean statements for the ternary operators here are simply repeated really. 

OR.. 
(can most definitely be a TODO for later but..) would it be possible to abstract the logic for the values to the specific language somewhere else? So that only _something like_ the below is possible with no `switch` statement needed? 
```
        scalarCapsLockKeyWidth = currentLanguage.scalarCapsLockKeyWidth 
        scalarReturnKeyWidth = currentLanguage.scalarReturnKeyWidth
        scalarSpecialKeysWidth = currentLanguage.scalarSpecialKeysWidth
        scalarReturnKeyWidth = currentLanguage.scalarReturnKeyWidth
```
Scribe-iOS isn't my domain, so let me know if I'm talking crazy :laughing:

Keyboards/KeyboardsBase/KeyboardViewController.swift
```diff
@@ -1596,20 +1597,21 @@ class KeyboardViewController: UIInputViewController {
...
         }
-        if DeviceType.isPad
+       /* if DeviceType.isPad
           && key == "$"
           && controllerLanguage == "Italian"
         {
           leftPadding = keyWidth / 3
           addPadding(to: stackView1, width: leftPadding, key: "$")
-        }
+        }*/
```
wkyoshida, line 1614: nit - We OK to just remove the newly commented code?

Keyboards/KeyboardsBase/KeyboardKeys.swift
andrewtavis: Either sounds good to me! What are your thoughts on this, @henrikth93? 

And all of Scribe is your domain, @wkyoshida 😊🤗

henrikth93: I think Will's idea of abstracting it is good, and that is something I can try and solve. @wkyoshida I am not sure I understand what you mean regarding the ternary operators, but do you mean that just checking the statement once would be less consuming? To tell the truth, my understanding of ternary operators and what they do under the hood is lacking.

henrikth93: @wkyoshida @andrewtavis I think the logic for setting the values could be moved to interfacevariables maybe

wkyoshida: Hey! :wave: 

> I am not sure I understand what you mean regarding the ternary operators, but do you mean that just checking the statement once would be less consuming?

Oh no, not really :smile: I just meant that since the boolean statements for them is the same, could just write it once, store the result in a variable, then consult the variable in each ternary. It's really just more of a visual suggestion to declutter the code a bit to make it more visually clean - that's all.

---

Also, regarding the abstraction - I'm good with leaving it for another PR, so no stress (unless you'd really like to tackle it here already, then by all means :smile:)

andrewtavis: Happy to move that to a new PR/issue or have it be committed here. @henrikth93, let me know which one you'd prefer and I'll get to the review soon if we're leaving as is 😊

henrikth93: Let us move that to a new issue! :)

Keyboards/KeyboardsBase/KeyboardViewController.swift
henrikth93: Yeah, It would be OK, I commented it out realizing there were better ways

andrewtavis:
af9e726 removes the code that was commented out and also adds a char to Spanish and Swedish in one row as they had less than the others. Big thing that I'm realizing is that the Russian keyboard is automatically crashing at this point, but then this is the case on main as well 🤦‍♂️ I'll take a look at it. @henrikth93, as a final commit for this, do you want to change it so that the right shift key takes up the whole of the extra space as is done on the system German keyboard? See below:

Simulator Screenshot - iPad Pro (12 9-inch) (6th generation) - 2024-02-06 at 23 52 54

I think that this would be the easiest way of figuring this all out 😊

andrewtavis:
Looking into this now and making a checklist for myself to confirm that the UI is ready for things:

- [ ] French letters
- [ ] French alt chars
- [ ] German letters
- [ ] German alt chars
- [ ] Italian letters
- [ ] Italian alt chars
- [ ] Portuguese letters
- [ ] Portuguese alt chars
- [ ] Russian letters
- [ ] Russian alt chars
- [ ] Spanish letters
- [ ] Spanish alt chars
- [ ] Swedish letters
- [ ] Swedish alt chars

andrewtavis (approved):
Approved and on to the next issue! 🥳🚀

63d5ce1 has all kinds of changes coming through:

- I simplified how expanded keyboards are laid out such that we always have the same number of keys per row for all languages
  - Just felt it was easier than going in and adding all the padding
  - We can do this later if there's a demand
  - This included some general resizing of the special keys
- There was a bug that brome the Russian keyboard where `ussianKeyboardConstants.symbolKeysPadExpanded` was assigned to `letterKeys` rather than `symbolKeys`
- I made the height of expanded keyboards a bit larger
- The vertical space between the keys on expanded keyboards was decreased
- I removed the menu option for putting a period and a comma on the letter keys if the user is on an iPad
- Lots of keys were added across all keyboards 😊

Thanks so much for your dedication in working on this, @henrikth93! Really so great to have this finalized 🙌

Issue #393, opened 2024-01-13
Add external link icons to About menu options


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something that's needed for the new Scribe menu is to have external link icons for those options on the `About` screen that link to the web. Options that need this icon are:

- See the code on GitHub
- Chat with the team on Matrix
- Share Scribe
- Rate Scribe
- Report a bug
- Send us an email

All of these options are not inherently links, but it would be good for consistency if each menu option has an icon.

### Contribution

Happy to work on this myself or support someone who has interest in picking it up!

kevin9foong:
hi @andrewtavis, could I find out more about this issue? 

![image](https://github.com/scribe-org/Scribe-iOS/assets/55353265/d28bbbac-7824-4fff-84cb-8734009215e6)
From the `About` screen, I can see that there are currently already icons on the left hand side of the text. 

Would you perhaps be referring to adding new icons that you'd like to place on each menu option?

andrewtavis:
Hey @kevin9foong 👋 Specifically for the options I mentioned above we'd like to have an icon like the [external link icon](https://de.m.wikipedia.org/wiki/Datei:External_link_font_awesome.svg) :) Would you have interest in working on this?

andrewtavis:
Closed by #398 🎉 Thank you, @Jag-Marcel! As you already picked up on #391 for your PR, would you want to do that as a quick chore and then we can find another issue for you that's a bit more challenging? You removed the name of the file when you added the license for the ones you sent along, which thinking about it that's likely not the best practice and maybe we want to change it so it's just the license as you did it? Your ideas would be welcome in that issue, and then as I said it would be wonderful to figure out something further for you to work on!

Jag-Marcel:
Thank you!
Sure, I can do that, but I have exams the next few weeks, so I might not be able to do much for the project at the moment. I'll get to it as soon as I have more time!

> You removed the name of the file when you added the license for the ones you sent along, which thinking about it that's likely not the best practice and maybe we want to change it so it's just the license as you did it?

[...]

PR #398, opened 2024-02-08
Added external link icon in "About" page (#393)

Jag-Marcel:

---

### Description

This adds an icon to show that several buttons in the "About" page in the app link to external sites, mainly the ones outlined in the issue #393, with the exception of the "Share" button, which doesn't necessarily link to outside the app. I also thought that the "Share" icon was already universal enough for people to interpret what the button does at a glance.

I used the icon that was used as an example in the original issue, but it can be replaced by similar ones that might fit the style of the rest of icons better if there is a need. 

I tested by using the Xcode iPhone 15 Pro, Pro Max and iPad Pro 11in simulators

### Related issue

- #393


andrewtavis:
Generally everything's looking great here, @Jag-Marcel :) One thing to note is that the resolution of the icon isn't scaling particularly well. I guess I'd also prefer if the arrow was going right out at 45 degrees rather than say 55. Would something like [this icon](https://icons8.com/icon/set/external-link/ios-glyphs) be possible. The other thought I'm having on this is also that the icon would ideally be grey like the chevrons that indicate navigating within the app.

Let me know if you'd like to work on the above two points! Thanks for the work till now 😊

andrewtavis:
I realized that designs were missing for this, @Jag-Marcel :) I did some minor edits for it just now, which can be seen on Figma here: https://www.figma.com/file/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?type=design&node-id=513%3A2206&mode=design&t=ReaZQG3S33ML5O2k-1.

Jag-Marcel:
Alright, I'll get to it. Should be a fairly simple change.

andrewtavis:
Nice @Jag-Marcel, thanks for looking into this further 😊

andrewtavis (approved):
This is really great for the first contribution, @Jag-Marcel 😊 Appreciate the quick turn around in addressing the comments. Would be great to work with you on another issue!

Will bring this in and do some minor edits I'm seeing that are unrelated before closing the issue :)

Issue #391, opened 2024-01-13
Add license information at the top of each Scribe file


andrewtavis:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

Based on a comment in #384, it looks like something we're missing for the GPL license is a notice at the top of each file. The directions for releasing GPL based code can be found [here](https://www.gnu.org/licenses/gpl-howto.en.html). It would be good for us to check what else is required and expand this issue to include other elements we're missing.

Happy to discuss this with people and implement parts of it once we've figured out all the changes that are needed!

Jag-Marcel:
Alright, currently my plan would be to add something like this to every file:
```
/**
 * [Short description
 * of the file]
 * 
 * Copyright (C) 2023-2024 Scribe
 * 
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
```
The short description would be something like "This file controls outlets to the buttons on the 'About' page of the app" (for AboutTableViewCell.swift); the rest is the same I found on other files that already had a license notice.

andrewtavis:
Sounds great to me, @Jag-Marcel! Appreciate you taking this on :) If you need any help with the file descriptions, then just let me know. Also happy to do some of those if you're not sure 😊

andrewtavis:
Closing this up 🥳 Thanks for all this, @Jag-Marcel!

PR #403, opened 2024-02-28
Added copyright notice to all .swift files (#391)

Jag-Marcel:
### Contributor checklist

---

### Description

Added a generic copyright notice to all .swift files in the directory, with a short description for each file. For those that already had a description, I just copied the existing one; for others, I came up with one to the best of my abilities.

### Related issue

- #391

### DIFF
--- a/Keyboards/InterfaceConstants.swift
+++ b/Keyboards/InterfaceConstants.swift
@@ -1,6 +1,5 @@
 /**
- * Constants for the Scribe keyboard interfaces
- *
+ * Constants for the Scribe keyboard interfaces.
  *
  * Copyright (C) 2023 Scribe
  *
--- a/Keyboards/InterfaceConstants.swift
+++ b/Keyboards/InterfaceConstants.swift
@@ -1,8 +1,22 @@
-//
-//  Scribe
-//
-//  Constants for Scribe keyboard interfaces.
-//
+/**
+ * Constants for the Scribe keyboard interfaces
+ *
+ *
+ * Copyright (C) 2023 Scribe
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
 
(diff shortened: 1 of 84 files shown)


Keyboards/InterfaceConstants.swift
```diff
@@ -1,8 +1,22 @@
...
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
```
wkyoshida, line 19: Hey!! :wave: 

Just wondering about something here.. 
How would we plan on maintaining this license header in all the relevant files? Not sure if some automation was used to make the changes in this PR, but could there be something like [.editorconfig](https://editorconfig.org/) that we could use to generate these (especially if we'd like to tweak the text in the future)?

CC: @andrewtavis

andrewtavis: Any level of automation would be appreciated for this. I believe that @Jag-Marcel was going through individually as this was also a task of better understanding the codebase. The file headers also have a description of the file, which we thought was important. Could you send along an example of how the license could be automated with this, @wkyoshida?

Keyboards/KeyboardsBase/Colors/UIColor+ScribeColors.swift
```diff
@@ -1,6 +1,22 @@
-//
-//  UIColor+ScribeColors.swift
-//
+/**
+ * Adds Scribe colors to the UIColor pool
```
andrewtavis, line 2: Minor change request on this, @Jag-Marcel, could we put a period at the end of all of the descriptions and just have one line after it instead of two?

Jag-Marcel: Sure! I'll get on it

Keyboards/InterfaceConstants.swift
wkyoshida: It seems like this might be possible with Xcode perhaps ([ref](https://stackoverflow.com/questions/69420028/incomplete-file-copyright-headers-in-swift-package-with-xcode-12-13))? I don't have Xcode to be able to validate if this works though..

andrewtavis: This seems to be a bit more for changing the default for when the file is made, but maybe I'm not reading it clearly 🤔 Would be good to include though! I'd say it's a different issue though.

andrewtavis (approved):
Just finished going through all these, @Jag-Marcel 🚀 Thanks so much for the hard work here and the quick edits 👏 Let me know if you have interest in working on #405, and aside from that we can find the next issue for you :)

Issue #361, opened 2023-09-18
Allow for swiping between conjugation and declension views


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would add swipe functionality to the conjugation and declension views for Scribe keyboards, with the former being displayed when a user presses the conjugate key and enters a verb and the latter being displayed when a user clicks on a preposition's case to see the possible pronouns that could be used. As of now we have a left and right button to the side of the view that are used to switch views, with the goal being that the user would also be able to swipe. The left and right buttons would not be removed to make sure that the user can still understand the ability to switch views :)

The code for this would generally follow the patterns used in #337 that added swipe support to the new menu 🚀

### Contribution

Happy to work on this with someone or get to it myself! 😊

andrewtavis:
CC @SaurabhJamadagni on this :)

andrewtavis:
CC @Taufi, can you write in here so I can assign you as well :)

andrewtavis:
Exactly, @Taufi :) As soon as you write in I can assign you. Helps us keep an overview of who's on what 😊

Let me know if there's anything I can do to help!

andrewtavis:
Many thanks for checking in, @Taufi!

henrikth93:
> Thanks for the research so far, both of you! 🙌 Both are really good points!
> 
> Originally I was typing out that `ViewDidLoad` makes sense to keep the keyboard load times down, but that's also with the assumption that a person will use conjugate at least once when using Scribe. Maybe for Scribe-Desktop where someone's writing an email or a paper, but I'd say that in a typical phone interaction a person is using verbs that they generally know how to conjugate, and just need help in special circumstances. In this case I'd go with `setConjugationBtns`, with this also having the architectural bonus that its usage is closer to its function :)
> 
> How does this sound?

Makes sense!

andrewtavis:
Closed by #412 🥳 Thanks so much, @Taufi and @henrikth93 for the great collaboration on this!

PR #412, opened 2024-04-12
Allow for swiping between conjugation and declension views #361

Taufi:
### Contributor checklist

---

### Description

The pull request provided by henrikth93 and Taufi offers a solution for issue #361: enabling swiping between conjugation and declension views. Alongside the existing left and right buttons used for view switching, users can now also swipe to navigate between views.

The fix is tested by henrikth93 and Taufi on

iPhone 12, iOS 17.3.1. (physical device)
iPhone 15, iOS 17.4. (simulator)
iPhone 15 Pro, iOS 17.0. (simulator)
iPad (10th generation), iOS 17.0 (simulator)
iPhone SE (3rd generation), iOS 17.0 (simulator)

Please note: Swipe gestures do not work with all simulators. It probably depends on whether XCode is running under Rosetta or not (https://forums.developer.apple.com/forums/thread/696832)

### Related issue

- #361

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -1207,13 +1207,13 @@ class KeyboardViewController: UIInputViewController {
 
   /// Sets up all buttons and labels for the conjugation view given constraints to determine the dimensions.
   func setConjugationBtns() {
-    // add swipe functionality to the conjugation and declension views
+    // Add swipe functionality to the conjugation and declension views.
     let swipeRight = UISwipeGestureRecognizer(target: self, action: #selector(shiftLeft))
     keyboardView.addGestureRecognizer(swipeRight)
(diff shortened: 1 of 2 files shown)


Keyboards/KeyboardsBase/KeyboardViewController.swift
```diff
@@ -1207,6 +1207,13 @@ class KeyboardViewController: UIInputViewController {
 
   /// Sets up all buttons and labels for the conjugation view given constraints to determine the dimensions.
   func setConjugationBtns() {
+    // add swipe functionality to the conjugation and declension views
```
andrewtavis, line 1210: Quick note here that we want comments on their own line to be complete sentences with a period at the end :) No stress, I'll fix it 😊

andrewtavis (approved):
Thank you both so much for this! Another great feature 😊 Looking forward to playing around with this when it's live!

Issue #390, opened 2023-12-28
Add linting checks to CI process


wkyoshida:
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue is for setting up linting checks on PRs as part of our CI process. Also related to CI:
- https://github.com/scribe-org/Scribe-iOS/issues/8
- https://github.com/scribe-org/Scribe-iOS/issues/326

Issue derived from [PR discussion](https://github.com/scribe-org/Scribe-iOS/pull/389#discussion_r1435749064)

wkyoshida:
Ideas could be (though open to discussion):
- to use [SwiftLint](https://github.com/realm/SwiftLint)
- for a GitHub action to display warnings on violations
- for a GitHub action to notify the contributor to manually fix violations and/or automatically fix them by running linting themselves when applicable

andrewtavis:
Thanks, @wkyoshida! I'll leave this for a moment, but can also get to it myself if it's not picked up :)

shashankiitbhu:
Hi @wkyoshida @andrewtavis can I work on this issue? Really Interested in this overall project and would love to start contributing to it by this

andrewtavis:
Makes sense to me! Feel free to do some research on this, @shashankiitbhu, and we can then discuss this a bit in the dev sync on Saturday 😊

andrewtavis:
Hey @shashankiitbhu 👋 Doing a quick check in on this issue :) Anything we can do to help?

henrytwagner:
Hello, I would definitely be interested on working on this if help is needed!

andrewtavis:
Currently on a call with @henrytwagner to discuss a potential issue that could be worked on for a class on debugging and quality assurance. This issue came to mind as something that would be fitting to work on 😊 I'll reassign this issue under the circumstances :)

@henrytwagner, as discussed, what it is that we're looking for here is:
- Find Swift repositories that have tests running on their PRs
- Find ones that have linting and/or formatting checks
- Check the `.yaml` configuration file for the tests that will be in the `.github/workflows` directory of the given project (the name of the test on the PR will be the same name as the file)
[...]

PR #417, opened 2024-04-26
Added linting check using SwiftLint

henrytwagner:
### Contributor checklist

---

### Description

Previously there was no linting on pull requests. Here I added two files: pr_swift_lint.yaml, which runs the linting check and displays a helpful output within the check output, and .swiftlint.yaml which is the configuration file for the linting done in pr_swift_lint.yaml and currently only alters the default configuration by allowing 1 character variable.

### Related issue

- #390


andrewtavis:
Thank you, @henrytwagner! This is definitely in the direction we need it to go in 😊 So happy to finally have the linting going in the right direction! I'll definitely need to do a squash the commits here, but overall very happy :) :) 

Will give this a more thorough look a bit later today!

.github/workflows/.swiftlint.yaml
```diff
@@ -0,0 +1,37 @@
+# SwiftLint Configuration File
```
andrewtavis, line 1: Really appreciate you adding all the config options in here and also documenting what will cause errors and warnings once they're turned on, @henrytwagner! Will really help us organize the work for this going forward 😊

andrewtavis: I'll likely move this file to the project root, but for the purposes of this PR it's totally fine to have it here. I figure this is where ESLint and Prettier configs normally are. Will send a commit with this along later :)

.github/workflows/pr_swift_lint.yaml
```diff
@@ -0,0 +1,84 @@
+name: pr_swift_lint
```
andrewtavis, line 1: For here, I'm expecting that there's a pre-prepared workflow that can be subbed in for parts of this, but is looking really good all the same :)

andrewtavis (approved):
Happy to bring this in and use it as a basis of what will come next, @henrytwagner! As stated, I'll do a squash of all the commits, and will send something along later that does some formatting and edits the settings a bit.

Would be happy to work on getting some of these linting errors/warnings fixed with you at some point. If you wanted to, we could also discuss making some `help wanted` or `good first issue` issues based on the outputs of the process.

Thank you for the contribution!

andrewtavis:
Hey @henrytwagner 👋 After doing some minor edits, we have a functioning linting check based on SwiftLint and the swiftlint.yml (renamed from yaml as it was working better). See #419 to check it out! Really appreciate the help here!

Issue #421, opened 2024-05-05
SwiftLint: Fix trailing comma throughout codebase


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue would be to remove `trailing_comma` as a disabled rule from [.swiftlint.yaml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yaml) and fix all errors for it in the codebase.

We'd know that this is successful if our SwiftLint pull request workflow passes on the pull request for the change :)

### Contribution

This would be a great `good first issue` for someone wanting to start with Scribe-iOS development. Happy to support! 😊

andrewtavis:
Closed via #422 🚀 Thanks, @mostazomarc! Would be very happy to work with you on more issues like this 😊

andrewtavis:
Hey @mostazomarc 👋 Tons of new SwiftLint issues just made 😊 Feel free to write in one and I'll assign you!

PR #420, opened 2024-05-03
#292 Load UI lexicon

Taufi:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

The UILexicon is unable to load due to the Language Database being located in the app's main bundle, where writing data is not permitted. The solution involves copying the database into the application support directory and accessing it from there.

Additionally, the pull request includes a minor fix for an issue in ParentTableCellModel.

The fix is tested on

- iPhone 12, iOS 17.3.1. (physical device)
- iPhone 15, iOS 17.0. (simulator)
- iPad (10th generation), iOS 17.0 (simulator)
- iPhone SE (3rd generation), iOS 17.0 (simulator)

The simulator tests are essentially retests because the error doesn't manifest there.

### Related issue

- #292

PR #422, opened 2024-05-05
#421 fix trailing comma

mostazomarc:
### Contributor checklist

---

### Description

Remove `trailing_comma` as a disabled rule from [.swiftlint.yaml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yaml) and fixed all errors for it in the codebase.

**The linter fix was tested running:** `swiftlint --no-cache --config ~/com.raywenderlich.swiftlint.yml`
**and tested on:**
- iPhone 15 Pro Max, iOS 17.4. (simulator)

### Related issue

- #421


mostazomarc:
The SwiftLint pull request workflow is checking for errors that it shouldn't be, right?

andrewtavis:
Seems like it, @mostazomarc :) There was another implementation of the workflow from #417, but then it made sense to me to just use the pre made action 🤔

Thanks for the contribution! Let's see if we can figure this out :)

andrewtavis:
Looking at the [workflow](https://github.com/marketplace/actions/github-action-for-swiftlint), seems like we're missing the paths options at the top? Do you want to add those and see if it works? Note that we have `.yaml` instead of `.yml`. Feel free to experiment with any of the other options you see that you think might help!

mostazomarc:
Sure! I'll go ahead and add the paths option at the top of the workflow file, I'll make sure to use .yaml instead of .yml.

mostazomarc:
Paths options didn't solve the problem.
I believe `norio-nomura/action-swiftlint@3.2.1` uses a config file named **.swiftlint.yml** (as it worked okay on #420) and it was renamed to [.swiftlint.yaml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yaml) in this [commit](https://github.com/scribe-org/Scribe-iOS/commit/81b1d1e987dd7d6d4d723995edeae2686a9f5eaa) so it no longer uses the configuration provided?
Maybe I could try renaming [.swiftlint.yaml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yaml) back to **.swiftlint.yml** or some workaround?

andrewtavis:
Trying to rename the file to `.yml` would be fine :) We can also rename the workflow to a `.yml` file as in the example. We normally use the longer version of the file extension, but it that's what's needed, then no stress at all 😊

andrewtavis:
So should be fine to just switch it back to `.yml` as you mentioned that it will work like #420. But maybe let's keep the path definition in there so it's a bit more verbose?

mostazomarc:
I actually don't know why is doing all these checks as they should be disabled by the `.swiftlint.yml` file.
I thought it was because of the `.yaml` file extension, but it seems not.

andrewtavis:
Ya I'm really not sure what's up... As a final check, let's maybe set it back to the state of the last passed tests via [the branch for it](https://github.com/Taufi/Scribe-iOS). Differences there are:

- The workflow is `.yaml`
- There are no paths in the workflow

Would you want to reset it so we can at least confirm that there's nothing new going on? Sorry for not being a bit more active on trying things. I'm currently traveling 🌴 Appreciate all the effort going into this!

mostazomarc:
Sure! I'll reset the last two commits while we figure out what's going on.

andrewtavis (approved):
Looks like we just need to keep the file as `swiftlint.yml`, @mostazomarc 😊 Appreciate the support here! Let me know if you have any interest in working on the other linting rules. Would be super happy to make some more issues for this, or we could also check in on making some issues together for yourself and others to work on!

Issue #423, opened 2024-05-16
Inconsistent command bar outline behavior


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Something that was discussed in #416 is that we're at times seeing inconsistent borders around the command bar. The following is from an iPhone 15 Pro simulator:

<img width="348" alt="image" src="https://github.com/scribe-org/Scribe-iOS/assets/63468878/1d979706-7d0e-44b3-b0c2-c2ce479554ad">

And the following is from an iPhone 15.2 Pro simulator:

<img width="463" alt="Screenshot 2024-05-07 at 3 19 09 PM" src="https://github.com/scribe-org/Scribe-iOS/assets/60264030/2cce3fb8-8f9c-491a-a9cf-bc75783f1250">

In the above, the shadows are darker under the buttons, and the border around the command bar is also darker. One way to improve this a bit would be to remove the border around the command bar on light mode, as the shadow is providing enough contrast (is needed on dark mode though).

Open to discussions on this and working with people on it!

### Device type

iPhone

### Versions

Current main version

andrewtavis:
CC @Jag-Marcel and @fabulouiOS-monk 😊 Let me know on the idea of removing the border around the command bar! Happy to get suggestions on what's up with the shadows as well :)

fabulouiOS-monk:
Hi @andrewtavis, I did some work on it and here are the Screenshots with the code updates:
**Removed shadow from below the bar**

In Dark mode:
<img width="387" alt="noShadow-Dark" src="https://github.com/scribe-org/Scribe-iOS/assets/60264030/402362a3-ea94-49f4-bf31-7f987aa078a1">

In Light mode:
<img width="385" alt="noShadow-Light" src="https://github.com/scribe-org/Scribe-iOS/assets/60264030/e4f7a2b8-2945-4324-a542-06c361d8d0ac">

fabulouiOS-monk:
If all good from your side I can raise  PR for it.

andrewtavis:
[...]

PR #438, opened 2024-05-18
Fabuloui os inconsistent command bar outline behavior 423 new

fabulouiOS-monk:
### Contributor checklist

---

### Related issue

- #423


andrewtavis (approved):
All's looking good to me, @fabulouiOS-monk! I added a minor change to remove the CommandBarBlend from the layout, which was a 1px wide element that was used to make sure that the width of the border between the Scribe key and command bar was only 1px instead of 2px with one coming from the border of each :)

Looking much more clean now!

PR #415, opened 2024-04-23
Added explanation for conjugations missing from Wikidata (#400)

Jag-Marcel:
### Contributor checklist

- [x] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

When a word is present in Wikidata, but doesn't have any data for conjugation, the buttons in the conjugation screen show "Not in Wikidata" in the specific keyboard's language instead.
Here I added a little info symbol to these buttons. Clicking them now leads the user to an info view describing how Scribe data is fetched from Wikidata, the same one used when entering a word missing from the database.

#### Example:
image
image

### Related issue

- #400

PR #448, opened 2024-05-29
Install screen version update

Jag-Marcel:

---

### Description

Changes the installation screen's description to use the new one from iOS 15.0 onwards, instead of from 17.2. Also changed the deployment target to 15.0 to reflect changes in this PR and in #415.


andrewtavis:
Quick check on this, @Jag-Marcel :) I remember we discussed this, but are we sure we want to increase the minimum version here? We could leave it at 13 and just use the old one from 15 on. The way that it's set up here, we have a minimum of 15 and check if 15 is available. I think keeping it at 13 might be best so that it can still be available on as many levels of devices as possible, but maybe I'm also forgetting another reason why we said 15.

Thanks so much for looking into this!

Jag-Marcel:
Ah @andrewtavis, I wasn't sure how it worked, I didn't know it would lock out older versions from downloading it. I don't think there was anything that necessarily required iOS 15 to function, but there were things like the buttons with the info icon that look much better on 15.
But a thing I thought of that I just checked is that there are actually no devices that support iOS 13 that don't also support iOS 15, so it wouldn't change much for compatibility. Do we want to change the parts with availability checks then?

andrewtavis:
This is some great investigating, @Jag-Marcel! Yes, let's remove the availability checks. Also feel free to remove the longer install directions, which is great as the new ones will be much easier to localize 😊

andrewtavis:
Hmmmm, but let me merge in the other PR and then you can bring in the changes on main into here? All kinds of conflicts if not. Will do that tonight!

andrewtavis:
Other PR and the changelog update for it are done, @Jag-Marcel :) Let me know if you have any questions here!

Jag-Marcel:
Changes are done, it's a lot simpler now in places! 
As an aside, should we think about using UIButton.Configuration in other places in the future? It's the implementation I used for #415's buttons with the info icon. It has a few benefits like resizing the text in buttons automatically and makes stuff like padding easier, but they wouldn't change much for the buttons we currently have and I don't know how much work converting it over would be

Scribe/Colors/ColorVariables.swift
```diff
@@ -23,30 +23,30 @@ import UIKit
 var scribeKeyIcon = UIImage(named: "scribeKeyIcon")
 
 // Initialize all colors.
-var keyColor = UIColor(.key)
-var keyCharColor = UIColor(.keyChar)
-var specialKeyColor = UIColor(.keySpecial)
-var keyPressedColor = UIColor(.keyPressed)
+var keyColor = UIColor(ScribeColor.key)
```
andrewtavis, line 26: Thanks for also going through and changing color assignments! I agree that it makes sense for us to be explicit with the source :)

andrewtavis (approved):
Thanks for handling all of this, @Jag-Marcel :) Really great to have the care put into this and allow us to move on from so many of the old version checks 😊

andrewtavis:
> Changes are done, it's a lot simpler now in places! As an aside, should we think about using UIButton.Configuration in other places in the future? It's the implementation I used for #415's buttons with the info icon. It has a few benefits like resizing the text in buttons automatically and makes stuff like padding easier, but they wouldn't change much for the buttons we currently have and I don't know how much work converting it over would be

And re the above: let's not focus on this in the short term, but you'd be welcome to make an issue for this :) Sounds like a medium good first issue :)

Jag-Marcel:
Will do 👍

Issue #392, opened 2024-01-13
Convert i18n over to a JSON directory


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Something that Scribe-iOS will need to implement for a common localization process with Scribe-Android is to have a common JSON directory that has all of the localized strings that are needed for the app. This directory would then be controlled by [Scribe-i18n](https://github.com/scribe-org/Scribe-i18n) in a way that changes to i18n would then trigger PRs in Scribe-iOS and Scribe-Android (and any other repos). We'll need a base `en-US.json` file and then a file for any other language that we're localizing Scribe into 😊

We're choosing to stick with JSON to assure that localization tools like [Transifex](https://www.transifex.com/) or [weblate](https://weblate.org/en/) can be used for community localizations. We have some texts like the privacy policy that need formatting and thus their own `.swift` files, but for individual like menu option labels these can be stored within a common file. The general idea would be to save them in a JSON and then make use of [SwiftyJSON](https://github.com/SwiftyJSON/SwiftyJSON) to load them in and access the key-value pairs.

Parts to this issue:
- [ ] See if there's a good extension for this that is well maintained
- [ ] Set up Transifex/Weblate/TranslateWiki for Scribe
- [ ] Explore commits from external source i18n software to i18n repo to this repo
~~- [ ] Switch over string references in the menu to use JSON keys for localized strings~~
  - We'll have new issues for this :)

### Contribution

Happy to discuss the implementation of this and work on it or support! The focus should be on first getting all the English texts into the system, and then we can move over to [Scribe-i18n](https://github.com/scribe-org/Scribe-i18n) to start adding in new languages that will then be triggered when the user changes their system language.

andrewtavis:
@Jag-Marcel, can you write in here so I can assign? :)

andrewtavis:
We can get started on this once #415 is merged in 🚀

Jag-Marcel:
👍

andrewtavis:
Thanks @Jag-Marcel! ☀️☀️

andrewtavis:
[...]

PR #452, opened 2024-06-03
Localization overhaul

Jag-Marcel:
### Contributor checklist

---

### Description

This PR changes the text in the app to be localized using Xcode's string catalog. This means the app localization will change automatically when the language is changed, without having to use long switch cases for every text.

### Related issue

- #392


andrewtavis:
Something to check on this, @Jag-Marcel: in the [24/5/2024 GSoC Sync](https://etherpad.wikimedia.org/p/scribe-dev-sync-2024-5-24) we'd talked about making sure that the Scribe-i18n files are not included in the build so it's not sent to the App Store. Would you be able to include that in here if it hasn't already been done? :) I was also just about to ask if you could include removing the `fr.Iproj` too, but I see that that's been done 😇 Thanks!

andrewtavis:
Hey @Jag-Marcel 👋 Sorry for the merge conflicts above, but hopefully they won't be too bad to get around. Just hit send on v3.0 a bit ago 🥳 Note further that in c6b11f3 the option to change the app's language was removed, so we'll need to revert some of the changes in there including uncommenting the code and also changing some of the indexes from `[0]` back to `[1]`. Should all be doable, and then we can quickly move to v3.1.0!

andrewtavis:
I went ahead and reverted the changes in c6b11f3, @Jag-Marcel :) How do you want to handle the merge conflicts? Happy to help with these if you'd like some support!

Jag-Marcel:
@andrewtavis conflicts have all been resolved, and it should be ready to merge. I don't know if the text for the tip card should be included in this PR, but if they should, they still need to be added to i18n.

andrewtavis:
Nice @Jag-Marcel! Let me do that so I can practice the new setup a bit :)

Scribe/SettingsTab/SettingsViewController.swift
```diff
@@ -102,7 +100,7 @@ final class SettingsViewController: UIViewController {
       parentTable.tableFooterView?.isHidden = false
     }
 
-    footerButton.setTitle("Install keyboards", for: .normal)
+    footerButton.setTitle("Install keyboards", for: .normal) // is this ever used?
```
andrewtavis, line 103: This button appears on the settings screen if the user doesn't have an installed keyboard, @Jag-Marcel :) I'll remove the comment 😊

andrewtavis (approved):
Thanks for this, @Jag-Marcel! Really awesome to have all of this set out for the next big step for the app 😊 Will Remove the comment as I mentioned :)

Issue #329, opened 2023-06-18
Add app hints to convey instruction/information


wkyoshida:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

**WHAT**

This issue is for app hints on each app page, i.e. overlaid text that provide quick instruction/information about a page (purpose of the page, what a user can do on the page, etc.) These would:
- Exist for each app page
- Automatically display the first time a user visits a page (will have to store this interaction history)
- Use a light bulb icon :bulb: on each page that can be clicked to display the app hint for the page again 

**EXAMPLE USES**
- Could be used to also prompt users to install keyboards if none are installed yet
- Within the keyboard settings, could display: “Here’s where you can change how each Scribe keyboard functions”

### Contribution

- App hints planned as part of GSoC work by @SaurabhJamadagni  
- Designs for the app hints will be made by @andrewtavis

andrewtavis:
Designs are all done at this point 😊 I didn't put it in the prototype as I felt like it's not necessary to see exactly how it would work, but here's an image for documentation as well as the component :)

![App Hint Tooltip](https://github.com/scribe-org/Scribe-iOS/assets/24387426/cb8af737-f132-49b6-a861-fb8d85a399d4)
![About Scribe - Light](https://github.com/scribe-org/Scribe-iOS/assets/24387426/be7e1971-c25e-4b25-bc0f-a883f8b554bf)

fabulouiOS-monk:
Hi @andrewtavis, What's the status of this? Has anyone picked it up yet? 
If not I can start with this 😄 !

andrewtavis:
Hey @fabulouiOS-monk 👋 Sounds good :) This would be great to get some help with! As a first step, would you document some of the steps you'll take to work on this? Then once we agree on things you can get started on the coding 😊

fabulouiOS-monk:
Hey @andrewtavis, Here's the Action plan:

1. Implement the UI for the tip card (let us call it a tip card). Since you already have the design prepared I will proceed with that, I just will need **icon**, and **Color coding**.

[...]

PR #459, opened 2024-06-10
Add app hints to convey instruction information 329

fabulouiOS-monk:
### Contributor checklist

---

### Related issue

- #329


fabulouiOS-monk:
Hey @andrewtavis, I had to make some changes in **KeyboardViewController.swift**, to fix swiftLint issues. I don't know how it was merged in the first place to the main.

andrewtavis:
Nice, @fabulouiOS-monk! Super great 😊 And just in time for the v3.0 release! 

Not sure how the SwiftLint errors got in either 🤔 The checks are only running on PRs, I think, so likely just me adding them in via commits 😅 Thanks for fixing them!

I'll check this later this evening :)

fabulouiOS-monk:
@andrewtavis, One more thing, I have only added it to the AboutViewController, We need to add it to other places wherever needed, similar to what we did here. 😃

andrewtavis:
Nice, @fabulouiOS-monk :) Sounds doable! Do we have the reset app hints option in the menu already? Haven't been able to bring it down locally 😊

fabulouiOS-monk:
No @andrewtavis, in the current push we don't have the reset app hints option. Every time the user will visit to that screen, we will show that TipCard.

andrewtavis:
Ah ok, but let's definitely add this option in then and have some external variable that gets reset by the menu option being clicked :) 

Really exciting progress here! Such a nice feature to round out the application 😊

fabulouiOS-monk:
@andrewtavis  ^^, for reference above comment.

andrewtavis:
Hey @fabulouiOS-monk 👋 1ee3a6c sends along some styling and functionality edits. Big thing is that we need three different ones such that when the user closes one, they all don't close. That's now been added, and generally I think it's progress, but we have a major bug where the tooltip is behind the header and thus can't be clicked. You had them lower than where they'd naturally be right at the top of the page, butt hen when it's moved up it's obscured by the title text and can't be clicked (I'm assuming this is because the view of the page title is blocking it).

Do you want to bring the most recent changes down and see if you can bring the tooltips forward? Screenshots below:

Working fine:

simulator_screenshot_C0743E4C-2142-4368-BE57-ED6AADAEBE5C

And these two aren't:

simulator_screenshot_6851AF09-86A0-423F-A845-6A7F9F80FE0D

simulator_screenshot_88FE4870-2410-428D-91A9-EEB2C5A7818F

Let me know if you have any thoughts!

andrewtavis:
Also, just another quick note of something it would be cool to fix: On the About tab the distance between "About" and "Community" is less than the distance between "Settings" and "App settings" 🤔 Maybe we could move the table view with "App settings" up a bit so the distance to "Settings" is the same as the counterparts on the About page?

fabulouiOS-monk:
@andrewtavis, I tried with a few things to bring the tipCardView to the front but it doesn't work. One way I can think about is creating a UIView which will present on top of the TableView and in that transparent UIView we can show the tipCardView. Any other way you can think of?

andrewtavis:
Ya I was working at it for a bit as well before deciding to check in, @fabulouiOS-monk :) I think your idea of the UIView over the whole TableView is a good one! Let's give that a try, and if it doesn't work let me know and we can go back to the drawing board/ask the community 😊

fabulouiOS-monk:
@andrewtavis, I tried applying the above idea, half of it works, but when resetting the value through the AboutViewController sections it does not update the `settingsTipCardState` hence UI breaks.

andrewtavis:
Do you want to send along the most recent changes and I'll take a look, @fabulouiOS-monk? :)

fabulouiOS-monk:
@andrewtavis, Yeah sure, I push the changes. It only helps to show the tipCard initially at the top of the table view. but the presentation doesn't work correctly.

andrewtavis:
Hmmmmm, ya this is a tough one as it basically disappears and you can't bring it back, and then for me it also gets added above the install one such that you have two tooltip cards at once. Could we try adding it to the storyboard such that it's above the other parts, @fabulouiOS-monk? So we have a View that's above the headers for each of the pages in question, and then we link it to a variable and add in the tooltips onto those. That way we start from their position on top of everything, and then we build from there?

If the above sounds ok, then maybe we should revert the most recent commit and start from the last one?

We'll get this!

fabulouiOS-monk:
@andrewtavis, Finally the above issue is solved 😄 . It was such a simple solution, instead of **view**, we needed to use the **navigationController.navigationBar**, since we want to show our tipCardView on top of the navigationBar and not the view. 

Took so many hits and tries 😅 . Congratulations!!

fabulouiOS-monk:
@andrewtavis, I wanted to join previous one too, but well god(my current organisation 😉) has different plans. I will try to join if no sudden plans are made.

andrewtavis (approved):
Amazing that this all got worked out, @fabulouiOS-monk! Really wonderful stuff here 😊 Looking forward to expanding these later when the menu has more options! 🥳

Issue #465, opened 2024-06-16
Clicking `OK` on app hint when it's over a menu option clicks the menu option


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

Some unexpected behavior for the app hints is that when they're over a an app settings option, then the settings option is clickable from beneath the hint. This is not common as it's uncommon for the settings pages to be scrollable at this time given how many options there are, but is the case for iPhones with home buttons like an iPhone 8 or SE. See screenshot below:

![Simulator Screenshot - iPhone SE (3rd generation) - 2024-06-16 at 12 37 28](https://github.com/scribe-org/Scribe-iOS/assets/24387426/19d39f3b-a239-4570-9e60-1436d90ad84d)

In this screenshot I've scrolled down slightly such that now the app hint is above the option to view the source code on GitHub. Clicking the app hint in this case opens the link to GitHub rather than closing the hint.

### Device type

iPhone

### Versions

v3.0 and current main

andrewtavis:
CC @fabulouiOS-monk :) Would you want to look into this? This and #462 would basically finish up the app hints, to my knowledge. We'd need to eventually figure out how to change them if say a user goes into a keyboard's settings or does an app tutorial, but for the current iteration these issues would be great to handle pre-v3.1! 😊

fabulouiOS-monk:
@andrewtavis, Ok let me check this out!

andrewtavis:
Thanks for this, @fabulouiOS-monk! :)

andrewtavis:
Closed by #468 😊 Thanks, @fabulouiOS-monk! 🚀

Issue #462, opened 2024-06-15
Add animation to app hints appearing and disappearing


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

In #329 we added in hints to the UI to explain what certain tabs are doing for the user. As of now these hints appear and disappear without an animation or transition. It would be great if they would include one! See below for an example of the app hints:

![Simulator Screenshot - iPhone 15 - 2024-06-15 at 12 51 08](https://github.com/scribe-org/Scribe-iOS/assets/24387426/4389a11b-c348-4f92-9ac5-b5299b0ad15a)

### Contribution

Happy to help with this or get to it myself eventually! Would also be great to get some support for this from the community 😊

andrewtavis:
@fabulouiOS-monk, do you want to write in here so I can assign as #468 will close this with the new work you have locally? :)

fabulouiOS-monk:
Yep, assign it to me. Thanks!

andrewtavis:
Closed by #468 😊 Thanks so much for the hard work here and the dedication, @fabulouiOS-monk! 🙏

PR #468, opened 2024-06-16
Clicking ok on app hint when its over a menu option clicks the menu option 465

fabulouiOS-monk:
### Contributor checklist

- [x]  This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description
Made the swiftUI view as userInteracable so that the table view's option doesn't get clicked on click of OK on app hints.

**NOTE:** _This is only initial commit to check whether it works or not._

### Related issue

- #465
- #462


andrewtavis:
As of now it's still not fixed, @fabulouiOS-monk. Here's a screen recoding from a 3rd gen iPhone SE simulator:

https://github.com/scribe-org/Scribe-iOS/assets/24387426/c0509d15-2ff8-4675-a16f-d1b7ac9d6919

Let me know if you have some thoughts here!

fabulouiOS-monk:
@andrewtavis, in the latest commit, pushed the changes for the blink animation.

andrewtavis:
Hey @fabulouiOS-monk 👋 This is really great! One final comment on this now that I'm able to play around with it: Can we have the tip view show up only if it's at the top of the screen? As of now the condition is that if it's scrolling up, then we show the tip, but then if we scroll down and then scroll up a bit, then it still covers some of the menu options. I think that a conditional that the current scroll position is the default would be the easiest to assure that the tooltip isn't over a menu option :)

Thanks for all the hard work here!

fabulouiOS-monk:
Ok, will see the fix for it, @andrewtavis!

fabulouiOS-monk:
@andrewtavis, the latest commit has the fix to show the tip card view only when the user is at the top. Please have a look 😃 !

andrewtavis:
Looking forward, @fabulouiOS-monk!

andrewtavis:
Hey @fabulouiOS-monk 👋 Made the speed that the tip view appears a bit faster, just from testing it out. Really wanted to bring it in, but then on testing with an iPhone 15 the functionality isn't working as the value of `private var topContentOffset: CGFloat = 116` seems to be only for iPhone SE, so it breaks the functionality and the tip disappears without coming back on larger devices. Where is the value of 116 coming from? Are we able to define this programmatically such that it'll be changed based on device dimensions, etc?

andrewtavis:
Also a note for myself, if we can figure out programmatically what the location of the About hint view is, then it'd be nice if we could also set that value to the height of `hostingController.view.frame` in Installation here:

https://github.com/scribe-org/Scribe-iOS/blob/d8663e5e4e189343a1fb242b8bfa894734eb09a1/Scribe/InstallationTab/InstallationVC.swift#L282

I'm realizing that the hint on the installation page is variable based on the height of the device, and the number we're looking for is likely similar for what to set 116 for in `topContentOffset` and 178 in Installation's `hostingController.view.frame`. Note that they don't seem to similar, but 116 is based on an iPhone SE, and I chose the value of 178 based on an iPhone 15 Pro Max. There's a good chance that both of these are being set to "What's vertical position of the About tip view relative to the top of the screen?" :)

Let me know on 116 and if you have any thoughts. Once again very close! After this we're done with tip views, I swear 😄

fabulouiOS-monk:
> Hey @fabulouiOS-monk 👋 Made the speed that the tip view appears a bit faster, just from testing it out. Really wanted to bring it in, but then on testing with an iPhone 15 the functionality isn't working as the value of `private var topContentOffset: CGFloat = 116` seems to be only for iPhone SE, so it breaks the functionality and the tip disappears without coming back on larger devices. Where is the value of 116 coming from? Are we able to define this programmatically such that it'll be changed based on device dimensions, etc?

Oops! Let me check it. 😅 I was only checking in SE and not others.

andrewtavis:
All good, @fabulouiOS-monk! Happens :)

fabulouiOS-monk:
@andrewtavis, pushed the fix for all types of devices. 😄 Hope it all works.

andrewtavis (approved):
All working perfectly, @fabulouiOS-monk! Thanks so much for taking this feature and really delivering through the entire thing. So great that the hints are in the state they're now in 👏 Looking forward to the next feature with you!

Issue #466, opened 2024-06-16
Increase text size on Installation tab for smaller devices


andrewtavis:
### Terms

- [X] I have searched [open and closed design issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Adesign)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

The installation screen text is a bit small in the current iteration of the app for smaller devices like iPhone 8 and iPhone SE - see the screenshot below:

![simulator_screenshot_A97EA77F-2D31-4765-9B8C-E2F8867C349A](https://github.com/scribe-org/Scribe-iOS/assets/24387426/36aee3ed-7a41-4471-87fc-48a688ca078f)

The screen below is quite large, so the size of the text being small isn't well reflected. Check this in a simulator to see! It would be great if we could increase this a bit and make sure that the installation screen text size is in line with the rest of the app screens for:
- iPhones with a button
- iPhones without a button
- iPads with a button
- iPads without a button

This would be great to include in v3.1!

### Contribution

Happy to support or work on this myself! This would also be a `good first issue` for someone with a bit of Swift experience 😊

andrewtavis:
@henrikth93: assigning you as discussed on Matrix 😊 Let me know if there's anything I can do to help!

andrewtavis:
Closed via #472 :) Thanks, @henrikth93! 🥳

PR #472, opened 2024-07-01
Checks for width of iPhone

henrikth93:
CHecking the width of the iPhone, this will then we bused to decide the font size.

### Contributor checklist

---

### Related issue

- #466


andrewtavis (approved):
Thanks for this, @henrikth93! As discussed, we can do any necessary fine tuning to this before the release :)

Issue #470, opened 2024-06-21
Add radio button menu option selection table


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As discussed in a recent call, we'd like to break #255 up into smaller issues :) This issue will focus on creating a menu table with radio buttons, which will be used in #255, #314 and others! Designs for this can be seen on Figma [here](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=544-1527&t=nLduZtGK7BTGyEgx-1) 🖼️

### Contribution

@Jag-Marcel will be doing this as a part of GSoC 2024 ☀️ Please write in here so we can assign, and let us know if any support is needed! 😊

Jag-Marcel:
Thank you 🙏

andrewtavis:
Closed by #476 🙌 So happy to have this done so we can use it for other menu options, @Jag-Marcel! 🚀

Issue #469, opened 2024-06-21
Allow menu options to have text to the left of the selector


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As discussed in a recent call, we'd like to break #255 up into smaller issues :) This issue will focus on adding an option to put text to the left of selectors in menu options, so when there's a chevron as the right icon of a menu option, we could then put text like the current translation language to its left. We'll also use this in #255 for the menu options themselves. Designs for various menu options with this element can be seen [here](https://www.figma.com/design/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?node-id=513-1616&t=nLduZtGK7BTGyEgx-1) 🖼️

### Contribution

@Jag-Marcel will be doing this as a part of GSoC 2024 ☀️ Please write in here so we can assign, and let us know if any support is needed! 😊

Jag-Marcel:
Thank you 🙏

andrewtavis:
Closed by #476 :) Thanks for the great work to make the menu so consistent, @Jag-Marcel! 🚀

Issue #458, opened 2024-06-07
Bring in cross language translation tables into Scribe-iOS


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Scribe-Data's translation processes have been expanded such that languages can be translated into many others, not simply English to the target language. Once the new files are ready, this issue would simply do the needed changes in this repo and Scribe-Data's to transfer them over and have them accessible from the Keyboard files so work on #307 😊

- Note that this is currently blocked by https://github.com/scribe-org/Scribe-Data/issues/70 and the creation of all of the given files/adding them into SQLite tables

### Contribution

Happy to work on this when it comes time or support on this! @Jag-Marcel, making this as a little mini issue just in case there's a bit more of a process here before you start working on #307 :)

andrewtavis:
Closing this up, as my recent commit to #476 brought in versions on the DBs that can be used for cross language translation. The rest of the work here will be covered in the Scribe-Data issues where the data processes will be ran in full :)

Issue #307, opened 2023-05-21
Switch base translation language based on menu selection


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

This issue is to implement the change to the base keyboard translation language that the user translates from. The related Scribe menu issue is #255. The designs for #255 are found in the [designs in Figma](https://www.figma.com/file/c8945w2iyoPYVhsqW7vRn6/scribe_public_designs?type=design&node-id=405-393&t=nCPFM9zwswgPWWI6-0) :) Once the user selects a new language for the given keyboard, they should then see this chosen language as their base language with the translation keyboard being for the base language.

Note that this issue is blocked by the following:
- #255
- https://github.com/scribe-org/Scribe-Data/issues/70
- #458 

### Contribution

Happy to support and work on this! This issue is relatively low priority for now as we need lots of other issues to be finished beforehand.

andrewtavis:
@Jag-Marcel, same here, can you write in so we can assign :)

Jag-Marcel:
Already got started on this a little, what's the status on the Scribe-data side of things? Do we have a system to translate between languages other than English?

andrewtavis:
Ya all of the individual files are ready to be used. The last step for this is https://github.com/scribe-org/Scribe-Data/issues/70 as it would be best to harness Google Collab and its GPUs for this. If there's no progress on that, I can also do the test, or also maybe this is something we could look into if you'd have interest :)

andrewtavis:
Closed by #476! Glad to have this all set up, @Jag-Marcel! 😊

PR #476, opened 2024-07-26
System to pick a translation language for each keyboard

Jag-Marcel:
### Contributor checklist

---

### Description

This PR creates a menu for each keyboard, where you can choose a language to be used for Scribe's translation feature. Previously, every keyboard would require English as the translation language.

### Related issues

- #255 
- #307 
- #469 
- #470 

### Blocked by

- #458
  - This PR needs to have the translation tables for it to support different translation languages
  - Currently, I chose `[language abbreviation]_translations` as the name for the tables, but could still be changed
- scribe-org/Scribe-Data#70
  - This also blocks #458, since it needs these tables to exist


andrewtavis:
Thanks for the detailed PR and fixing the merge conflicts, @Jag-Marcel!

andrewtavis:
Brought in the localizations to main so that we can get v3.1 out, @Jag-Marcel 😊 Would be great if we can get a "Follow us on Mastodon" localization key in! I'll bring in everything else and send along v3.1 now though as I'd like to get it out today. No stress on one menu option not being fully localized :)

andrewtavis:
@Jag-Marcel, the most recent commit adds in a table for each language with a `word` column and then ISO-2 code columns for each of the respective translation columns for the given language. There are a ton of weird conflicts here now... Let me know if you have thoughts on this would would like me to dive deeper into then! Generally I'd say we're keeping the `.sqlite` files that are in the branch and then we can update them again with the new data once I've ran all of the translation scripts. We should likely get rid of the changes to `Scribe.xcodeproj/project.pbxproj` and reimplement anything that's here as the changes to add in the English keyboard are likely more involved than the changes for the branch. I'm not sure on `Keyboards/KeyboardsBase/Utilities.swift`, but this is likely just some simple conflicts that can be resolved in the editor 😊 Happy to do a call to look into this!

Jag-Marcel:
Should be done as soon as we have the full tables now!

andrewtavis:
Nice, @Jag-Marcel! This is amazing stuff here :) I'll do a review as soon as I can. Let me know if there's anything else you'd have interest in looking into over the last week 😊

andrewtavis:
@Jag-Marcel, playing around with it a bit, and let me just say that you really have a knack for user experience :) My thought was always that the user would select a new radio button and then go back to the settings via the back button, but then given the UI they will see the change in the interface option, so it makes total sense that we take them back directly. As with the commands on highlighted words, well done 👏

andrewtavis:
We do have some keyboard key spacing issues that are just the damn conditionals for language getting mixed as always. Might be something that we should consider setting up an enum for at some point, but as you can see there's a space being applied after the `L` that's bringing all the keys in to close with the gap to `Ñ`:

Simulator Screenshot - iPhone 15 Pro Max - 2024-08-19 at 20 21 35

This is the German keyboard with Spanish translation language :)

Padding code in general starts [here](https://github.com/scribe-org/Scribe-iOS/blob/39d35d65258aa35c5ccf5de54f925861b33754e7/Keyboards/KeyboardsBase/KeyboardViewController.swift#L1706). 

For the German keyboard there are spacing issues for:
- French (padding issues)
- Russian (see image below)
- Spanish (padding issues)
- Swedish (padding issues)

Simulator Screenshot - iPhone 15 Pro Max - 2024-08-19 at 20 26 51

The issue above is more key size than padding, which is in [KeyboardKeys.swift#L212](https://github.com/scribe-org/Scribe-iOS/blob/39d35d65258aa35c5ccf5de54f925861b33754e7/Keyboards/KeyboardsBase/KeyboardKeys.swift#L212). Might need to check these sorts of things for iPads as well.

Any thoughts on this? An enum would be ideal, or maybe a variable for the translation language itself so that we're not checking between what the keyboard language is and whether translation is in use (not what the translation language actually is).

andrewtavis:
I'm also a bit confused on the order of the translations. When I'm translating from Spanish -> German I'd expect to be able to do `Brasil` (a `word` that we did transfer that's in `ESLanguageData`) to Brasilien (the `de` key), but I'm getting `Nicht in Wikidata` like it's not a key. Are we using `DELanguageData` for this? I don't get `Nicht in Wikidata` for trying to translate `00` that's a `word` in `DELanguageData` which returns back `00` as it is in its `en` key, but then this should not be found in this case as it's not in `ESLanguageData`.

Ah I see what might be going on here (pardon the stream of conscious). We have it setup where it's from `es` in `DELanguageData` to `word`? It should be from `word` in `ESLanguageData` to `de` :) Reason for this is that we want there to be a unique result for any word that we'd return. In the machine translation process there are multiple words that might have been/might be translated to the same word, and then if we go from `es` to `word` in `DELanguageData` we'd then have multiple `word` options to choose from. Thinking of the Spanish word for over, `sobre`. As of now we've had `'nüber` translate to it, and we can assume that `über` will to, but we wouldn't want to return the former. We want to translate `sobre` to `über`, have `sobre` be the word and `über` be the `de` key, and then have a unique result for each word that we're translating :)

Let me know if this makes sense! Hope I'm understanding what I'm seeing from the translation results :) Will look into the code a bit more later.

Jag-Marcel:
I'll look into the first issue, maybe a function that returns the controller language's translation language is good? That could also simplify other places where it's used.
I'll change the way the translation tables are used so it works as intended as well.

Jag-Marcel:
> maybe a function that returns the controller language's translation language is good?

Turns out that function already exists, I just forgot I made it at some point 🙃

andrewtavis:
Looking forward to the changes, @Jag-Marcel! :)

andrewtavis:
Hmmmm, ya this could prove to be a problem. Maybe the structure of the database is off? I understand your original inclination on this to use one DB for the whole thing rather than go across multiple, which also makes sense given downloading data in the future. As I'm thinking, we'd now need to bring down the `ESLanguageData` DB to translate to German, which is a bit weird, but I guess does make sense in so far as if the user has say German and French, then they'd just have the `ESLanguageData` for both :)

I'll take a look at this later!

How are we on the layouts?

Thanks for all the efforts, @Jag-Marcel!

Jag-Marcel:
The layouts should be done, I'll try looking into the DBs some more as well.
Maybe they don't get initialized quick enough for data to be read because of some async processes? We could also try putting translations for all languages in one separate database file

andrewtavis:
That's a thought as well, @Jag-Marcel! Could be something in the language keyboards base where we have all the tables in one names after their source language or something like that. If we can't figure out the current implementation, then I can rework the process :)

andrewtavis:
Very sorry for the wait on all this, @Jag-Marcel! Thanks so much for your patience :) d4e2b0a sends along new language DBs without a translation table, and then there's a common `TranslationData.sqlite` DB that we can now use for translation. This will be consistent even post Outreachy, and we won't need to worry about a single language DB being assigned anymore.

How do you want to proceed here? Would you want to finalize the new process via the SQLite access of the new `TranslateData.sqlite` DB? Anything I can do to support?

Jag-Marcel:
@andrewtavis I'll look into this and tell you if I need anything, glad that the process and everything is finally figured out!

andrewtavis:
Ok we're good to go with the `.xcstrings` file, @Jag-Marcel 😊 I guess that I just didn't have Korean locally, which is what came down with the `git pull --rebase --autostash` :) See you tomorrow!

Jag-Marcel:
The localization not working was because we still had the old string keys in iOS, but I also had to disable the indents in the third party licenses for now because of the way we generated them. Could probably be reactivated in a future issue but do we even want to have them? I think the lists there looks fine even without indents.
Also we probably want to find a way to detect links in texts like the privacy policy and licenses. I refactored the way we add them here so we only need to add each link to a list once, but ideally we should have a way to find them from specific texts automatically.

andrewtavis:
Ah totally true, @Jag-Marcel! This PR then also closes #503 🚀 Thank you!

andrewtavis (approved):
So happy to have this merged, @Jag-Marcel! Thanks so much for the hard work and determination to see this through to the end 😊 The implementation exceeds expectations given its everything we were looking for + your improvements to the user flow with the language being selected directly. Such an amazing step towards making Scribe an accessible application to language learners from all over the world 🌐

Issue #477, opened 2024-07-26
Remove accent characters menu option from iPads


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

An issue that we have right now is that the disable accent characters option is available for iPads when it shouldn't be. Enabling this option crashes the keyboard, so it should be removed as this option doesn't add much for an iPad user (likely wouldn't be used).

One consideration though is that the Layout section of the menu would then be empty. We thus should conditionally render the settings section if there is an option present, or remove the entire section if it's not needed :)

Happy to support or work on this!

### Device type

iPad

### Versions

v3.0 and current main

fabulouiOS-monk:
@andrewtavis, you can assign me this now. I will be in touch with you regarding the details. Thank you!

andrewtavis:
Thanks so much for picking this up, @fabulouiOS-monk! 🚀

andrewtavis:
Closed by #480! 🎉 Thanks so much for the quick work here, @fabulouiOS-monk! 🚀

PR #480, opened 2024-07-29
Remove accent characters menu option from i pads 477

fabulouiOS-monk:
### Contributor checklist

---

### Description

Implemented a check for iPad to remove **Layout** section from langSpecific settings.

### Related issue

- #477

### DIFF
--- a/Keyboards/KeyboardsBase/InterfaceVariables.swift
+++ b/Keyboards/KeyboardsBase/InterfaceVariables.swift
@@ -22,7 +22,7 @@ import UIKit
 // A proxy into which text is typed.
 var proxy: UITextDocumentProxy!
 
-// MARK: - Display Variables
+// MARK: Display Variables
 
 // Variables for the keyboard and its appearance.
 var keyboard = [[String]]()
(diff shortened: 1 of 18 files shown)


fabulouiOS-monk:
@andrewtavis Please review PR, and check whether this is what you wanted.

andrewtavis:
Hey @fabulouiOS-monk 👋 Checking on this: Are we checking if there are no menu options and then removing it if so, or just removing it for iPads? The latter works well for now, but there will be options for this for iPads in the future, so the best thing would be to remove the option from iPads - this dropping the total options to 0 - and then write something that automatically removes it :)

fabulouiOS-monk:
@andrewtavis currently we are removing the whole section if device in question is iPAD.

fabulouiOS-monk:
@andrewtavis, Added condition to check which menu option it is based on **sectionState**, and if there are no section we will remove the whole **Layout**.

andrewtavis:
Just sent along a basic file reformat and also changed `MARK: -` to `MARK:` throughout the codebase as marks work without the dashes, so I don't see why we'd keep them 🤷‍♂️😊

andrewtavis (approved):
Thanks so much for pulling this through, @fabulouiOS-monk! Am going to do the data work now and send off v3.1 as soon as I can 🚀 So amazing to get the quick and effective help to get the new version out!

PR #482, opened 2024-08-01
i18n additions

Jag-Marcel:
### Contributor checklist

---

### Description

Adds in new localizations from Scribe-i18n and fixes some code that messed up the files through Xcode's automatic generation

Xcode also orders the languages for a specific key alphabetically, currently English is always the last. This needs to be changed within Scribe-i18n's conversion scripts though. It also puts the state for any language names that we don't have localizations for as "stale", even if they're being used within the app. A fix for this would be to create variables for the names of all languages, but these would pretty much never get used, opinions on this @andrewtavis @henrikth93?


Scribe/SettingsTab/SettingsTableData.swift
```diff
@@ -92,12 +92,13 @@ enum SettingsTableData {
...
     var sections = [Section]()
 
-    for language in installedKeyboards {
-      guard let abbreviation = languagesAbbrDict[language] else {
-        fatalError("Abbreviation not found for language: \(language)")
+    for lang in installedKeyboards {
+      guard let abbreviation = languagesAbbrDict[lang] else {
+        fatalError("Abbreviation not found for language: \(lang)")
       }
+      let key = "_global.\(lang.lowercased())"
```
andrewtavis, line 99: Nice, @Jag-Marcel! Thanks for this :)

andrewtavis:
> It also puts the state for any language names that we don't have localizations for as "stale", even if they're being used within the app. A fix for this would be to create variables for the names of all languages, but these would pretty much never get used, opinions on this @andrewtavis @henrikth93?

We can certainly just put the variables for now to get rid of the `stale` declarations, if need be. Hopefully they all will be used at one point or another 🤔

andrewtavis (approved):
Thanks for the quick update here, @Jag-Marcel! Really great that we can get this out so quickly 🚀👏

Issue #141, opened 2022-03-01
Rework Scribe key press when selecting a word


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Many have commented recently that the functionality of the Scribe key when a word is selected could be reworked to present the commands and then run them over the selected word. What this means is that the command prompts would not be shown and instead the selected word would directly have the command ran over it. This would replace the current functionality where for nouns and pronouns have their form and coinciding case shown respectively.

A way to include the current functionality would be to include it as a selectable option instead of commands that aren't applicable. Specifically words that are already in the keyboard's target language would not need to be translated, so this command could be replaced. Commands that could be displayed given the type of word selected include:

Nouns: form, plural
Verbs: conjugate (and case once #81 is finished)
Prepositions: case

This feature could also later present the `Synonym` command when #20 has been finished. A `Definition` command could be included as well, with maybe the information from `Form` and `Case` being included so that only three options would be needed.

### Contribution

I need to get more feedback on this, but will implement it if it makes sense. Help would be appreciated if people have interest!

andrewtavis:
Something that could also be done for this is that the command options could be shown along with the annotation. The annotation could be shown over the translate button field if there is one, and aside from that the conjugate button and plural button could be shown. It doesn't seem reasonable that a person would want to do an autosuggestion at this time, so we could present the two other buttons with it and use the selected word as a source of the command when it's pressed. This would actually be quite simple to implement! 😊

Jag-Marcel:
Working on this one ✋

andrewtavis:
Closed by #485 :) Thanks so much for the great work here, @Jag-Marcel! 🚀

PR #485, opened 2024-08-03
Enter selected text into Scribe

Jag-Marcel:
### Contributor checklist

---

### Description

These are changes to the behaviour of the Scribe key when text in the text field is selected. Currently, it only shows you annotations for the selected word (i.e. "N" for neuter words). Here, I changed it so the selected Scribe command's action is executed on the highlighted text.
#### Before
Screenshot 2024-08-03 at 12 48 26

#### After (after pressing "Plural" in the command bar)
image

We also considered an option where the text would simply be entered into the command bar's text field, but decided against it:
Screenshot 2024-08-03 at 13 04 27

### Related issues

- #141

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -2199,32 +2199,23 @@ class KeyboardViewController: UIInputViewController {
 
     switch originalKey {
     case "Scribe":
-      if proxy.selectedText != nil && [.idle, .selectCommand, .alreadyPlural, .invalid].contains(commandState) { // annotate word
-        if [.selectCommand, .alreadyPlural, .invalid].contains(commandState) {
-          commandState = .idle
-        }
-        emojisToShow = .zero
-        loadKeys()
-        selectedWordAnnotation(KVC: self)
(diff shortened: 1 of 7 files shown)


andrewtavis:
A few things to consider here, @Jag-Marcel:

- What happens if the word isn't the last one in the proxy?
  - Do we replace the word where the other one is
- We also need to make sure that we're not including a space at the end of such commands that might cause an inadvertent double space for the user
  - Maybe we can check to see if there is a space after the word already so that we only add a space if there isn't one already 🤔

This is amazing! Thank you! 🚀

andrewtavis:
Talking about it in the sync: The general thinking is that the command should be directly ran :)

andrewtavis:
This is ready for a file review at this point, right @Jag-Marcel? :)

Jag-Marcel:
> This is ready for a file review at this point, right @Jag-Marcel? :)

Not yet, I still haven't changed the behaviour

andrewtavis (approved):
Really amazing work here, @Jag-Marcel! 😊 Thanks so much for putting so much care into this issue! Really has turned out wonderfully, and using it even in testing was really fun :D

Issue #425, opened 2024-05-16
SwiftLint: Fix orphaned_doc_comment throughout codebase


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Work for this issue would remove `orphaned_doc_comment` as a disabled rule from [.swiftlint.yml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yml) and fix all errors for it in the codebase.

Docs for `orphaned_doc_comment` are here: https://realm.github.io/SwiftLint/orphaned_doc_comment.html

We'd know that this is successful if our SwiftLint pull request workflow passes on the pull request for the change 🚀

### Contribution

This would be a great `good first issue` for someone wanting to start with Scribe-iOS development. Happy to support! 😊

jon-tous:
Hi, happy to take this on as my first issue here :) @andrewtavis thanks for the warm welcome!

andrewtavis:
Thanks for your interest in helping out, @jon-tous!

andrewtavis:
Closed by #487 😊 Appreciate the quick contribution and the great communication, @jon-tous!

PR #487, opened 2024-08-10
Enable SwiftLint orphaned_doc_comment rule and fix violation

jon-tous:
### Contributor checklist

---

### Description

Removed orphaned_doc_comment as a disabled rule from [.swiftlint.yml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yml) and fixed all errors for it in the codebase (there was only one violation).

After removing `orphaned_doc_comment` from `disabled_rules`, I fixed the only violation by removing an orphaned comment that was a duplicate of part of the file's header doc string. At that point, no SwiftLint warnings were shown, and running `swiftlint --strict` in the project root confirmed: `Done linting! Found 0 violations, 0 serious in 86 files.` This rule would now be enforced by SwiftLint to prevent further violations from entering the codebase.

Since only .swiftlint.yml and comments were changed, it would not affect any running code, however I verified on the iPhone 15 Pro simulator (iOS 17.5) that the app builds and runs as usual following this change!

### Related issue

- #425

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -12,7 +12,6 @@ disabled_rules:
   - type_body_length
   # WARNINGS IN PROJECT CURRENTLY
   - todo
-  - orphaned_doc_comment
   - blanket_disable_command
   - for_where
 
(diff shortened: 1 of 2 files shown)


andrewtavis (approved):
Fantastic, @jon-tous! We'd be super happy to have you work on a few more of the SwiftLint issues, or also let us know if there are other issues you have interest in!

Issue #428, opened 2024-05-16
SwiftLint: Fix for_where throughout codebase


andrewtavis:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

Work for this issue would remove `for_where` as a disabled rule from [.swiftlint.yml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yml) and fix all errors for it in the codebase.

Docs for `for_where` are here: https://realm.github.io/SwiftLint/for_where.html

We'd know that this is successful if our SwiftLint pull request workflow passes on the pull request for the change 🚀

### Contribution

This would be a great `good first issue` for someone wanting to start with Scribe-iOS development. Happy to support! 😊

jon-tous:
Happy to take this one on!

andrewtavis:
Assigned, @jon-tous! :) Let us know if there's anything we can do to help 😊

andrewtavis:
Closed via #488 🚀 Please let us know if there's another issue you'd be interested to work on, @jon-tous! Happy to also answer any questions you have about the project, with a call being possible as well if you'd find it helpful :)

PR #488, opened 2024-08-10
refactor(swiftlint): enable for_where and fix violations

jon-tous:
### Contributor checklist

- [X] This pull request is on a [separate branch](https://docs.github.com/en/get-started/quickstart/github-flow) and not the main branch

---

### Description

Removed for_where as a disabled rule from [.swiftlint.yml](https://github.com/scribe-org/Scribe-iOS/blob/main/.swiftlint.yml) and fixed all errors for it in the codebase.

After removing `for_where` from `disabled_rules`, I fixed all 6 violations by removing the if statement within each affected for-loop, and moving the condition into a `where` clause on the for-loop itself. This is the preferred syntax for SwiftLint enforced by the `for_where` rule, and this refactor does not affect functionality. After making those changes, no SwiftLint warnings were shown, and running `swiftlint --strict` in the project root confirmed: `Done linting! Found 0 violations, 0 serious in 86 files.` This rule would now be enforced by SwiftLint to prevent further violations from entering the codebase.

I verified on the iPhone 15 Pro simulator and iPad Pro 13-inch simulator (both iOS 17.5) that the app builds and runs as usual following this change and the affected constraints look as normal. I verified with one of these fixed violations that this refactor does not affect which elements run the body of loop/conditional, using print statements and comparing the values of all the possible constraints before vs. after the refactor.

### Related issue

- #428

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -13,7 +13,6 @@ disabled_rules:
   # WARNINGS IN PROJECT CURRENTLY
   - todo
   - blanket_disable_command
-  - for_where
 
 identifier_name:
   min_length: 1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Emojis for the following are chosen based on [gitmoji](https://gitmoji.dev/).
 
 # [Upcoming] Scribe-iOS 3.2.0
 
+### ♻️ Code Refactoring
+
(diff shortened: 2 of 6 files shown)


andrewtavis (approved):
Thanks so much for the quick work here, @jon-tous! As said after the last one, please let us know if there are other issues you'd have interest in working on! 😊

Issue #238, opened 2022-10-20
Allow for currency symbol selection


wkyoshida:
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Description

As touched on in this [discussion](https://github.com/scribe-org/Scribe-iOS/issues/229#issuecomment-1281076612), the option to select a currency symbol to use for a keyboard could be a good customization feature.

Some initial thoughts:
- Scribe could quickly prompt the user for the currency symbol during a keyboard setup. This could be done potentially as an addition to the walk-through being worked on in #130, i.e. this issue might be blocked by it.   
- Scribe could allow for separate currency selections per keyboard configured, e.g. a user could select `£` in the German keyboard and `¥` in the Spanish keyboard.
- From [here](https://github.com/scribe-org/Scribe-iOS/issues/229#issuecomment-1284372347), concerning the below:
  > if we do give them the option to choose on data download, then there needs to be a preset value so that we don't lose people in the download funnel 

  Since different currencies are sometimes used in countries where the same language is spoken, to not show "favoritism", Scribe could have the preset value for the currency be a neutral one, e.g. the `£` could be used for Spanish instead of the euro or the peso(s). This is a minor detail, and I'm not even sure if it actually would be welcomed by users (it might not), but just a thought.  
- Apart from the keyboard setup mentioned in the first point, an option within Scribe settings is probably a good idea in case the user wants to change the currency symbol for their existing Scribe keyboard. This is likely a separate issue as a `Settings` page would need to exist.

### Contribution

Willing to help with design and development :rocket:

andrewtavis:
Thanks for making this, @wkyoshida!

- Currently blocked by #16, but will be high on the list to implement once that's done 😊

andrewtavis:
Just assigned to @SaurabhJamadagni 🥳

- Note that this issue is blocked by the menu item issue in #314

andrewtavis:
@vickcoo, some quick notes here for if you'd like to work on this:

[...]

PR #492, opened 2024-08-22
Make the keyboard able to replace currency symbols

vickcoo:
the iPhone keyboard could be replaced with difference currency symbols

### Contributor checklist

---

### Description
In this task, modify the keyboard generation from static to dynamic using the builder pattern which allows the keyboard to replace currency symbols (also can replace other symbols).

### Related issue
* #238

### DIFF
--- a/Keyboards/KeyboardBuilder.swift
+++ b/Keyboards/KeyboardBuilder.swift
@@ -1,0 +1,56 @@
+/**
+ * Building a keyboard layout using a builder pattern
+ *
+ * Copyright (C) 2024 Scribe
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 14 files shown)


andrewtavis:
Hey @vickcoo! This is really great! 🚀 Seems like a really great way to go about this :) You'd be very welcome to update each of the language keyboard interface files with this method of building the keyboards. Remember that we also need to do the same for `genPadNumberKeys` and `genPadSymbolKeys` as well 😊

Thanks so much for getting to this so quickly and bringing such a great improvement to the project! 🙌

vickcoo:
Hi @andrewtavis, thanks for reminding me about the modification I missed, I've also finished the keyboards for each language.

andrewtavis (approved):
This is some great progress here, @vickcoo! My recent commit made the needed adjustments to the keyboards that we're working on, but aren't launched yet :)

Thanks so much for the ideation of the issue and for the great implementation! 🚀🙏

Issue #517, opened 2024-11-30
Tab bar section labels are not updating their language


andrewtavis:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

The labels of the tab bar of the application are not being updated properly when the user changes their localization. An example is below for Arabic:

![Screenshot 2024-11-30 at 14 14 08](https://github.com/user-attachments/assets/1b0e3b0c-601a-4a17-b7eb-e2cf0d5db9f2)

When the user changes localization, all labels in the tab bar are still English. When they change to the Settings and About pages, these labels do update to their correct localized labels, but then the same is not true for the Installation tab. Ideally all tab labels would be updated when the user changes their localization.

### Device type

iPhone and iPad

### Versions

Current development branch

andrewtavis:
Closed by #522 🚀 Thanks for another great contribution, @victorsaccucci! Please let us know if you'd have interest in another issue 😊 Happy to also discuss the project at some point if you'd like :)

PR #522, opened 2024-12-12
Fix: Update tab bar titles on language change

victorsaccucci:

---

### Description

Added logic to dynamically update the "Installation" tab bar item title when the language changes. The solution involved adding the following line to the InstallationViewController:

``` 
title = NSLocalizedString("app.installation.title", value: "Installation", comment: "")
```

arabic

### Observation
I forgot to remove the following code from SettingsViewController, which can be safely removed without any issues:

```
@objc func setText() {
    tabBarController?.tabBar.items?[0].title = NSLocalizedString("Installation", comment: "").uppercased()
}

override func viewWillDisappear(_ animated: Bool) {
    super.viewWillDisappear(animated)
    NotificationCenter.default.removeObserver(self)
}
```

These were added while I was working on solving the tab bar title issue, and they are no longer needed.

- #517

### DIFF
--- a/Scribe/SettingsTab/SettingsViewController.swift
+++ b/Scribe/SettingsTab/SettingsViewController.swift
@@ -55,7 +55,6 @@ final class SettingsViewController: UIViewController {
     setHeaderHeight()
     showTipCardView()
 
-    title = NSLocalizedString("app.settings.title", value: "Settings", comment: "")
     navigationItem.backButtonTitle = title
 
     parentTable.register(
--- a/Scribe/AboutTab/AboutViewController.swift
+++ b/Scribe/AboutTab/AboutViewController.swift
@@ -38,9 +38,7 @@ final class AboutViewController: BaseTableViewController {
 
   override func viewDidLoad() {
     super.viewDidLoad()
-
     showTipCardView()
-    title = NSLocalizedString("app.about.title", value: "About", comment: "")
 
     tableView.register(
       UINib(nibName: "AboutTableViewCell", bundle: nil),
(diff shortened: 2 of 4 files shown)


Scribe/SettingsTab/SettingsViewController.swift
```diff
@@ -82,6 +82,16 @@ final class SettingsViewController: UIViewController {
...
   }
 
+  @objc func setText() {
+      tabBarController?.tabBar.items?[0].title = NSLocalizedString("Installation", comment: "").uppercased()
+  }
+
+  override func viewWillDisappear(_ animated: Bool) {
+      super.viewWillDisappear(animated)
+      removeObserver(object: NotificationCenter.default)
+      NotificationCenter.default.removeObserver(self)
```
andrewtavis, line 92: Would you be able to take a look at this warning here, @victorsaccucci? :)

andrewtavis: Note that the `removeObserver(...)` was removed in 023d8e2.

andrewtavis:
Functionality wise this is definitely going in the right direction, @victorsaccucci 😊 Do you have an idea for why we have to switch to the tab in order to get the tab bar label to be localized though? Ideally we'd change the language and everything would be in the given language.

victorsaccucci:
I created this viewWillDisappear method inside the SettingsViewController class, and if you check the original source code before my push, it’s not there. So, since it doesn’t make any difference to the app’s functionality, it can be removed.

Now I pulled from the original repository into my branch and only added the following code:

```
title = NSLocalizedString("app.about.title", value: "Installation", comment: "")
```

andrewtavis (approved):
What a saga :) Thanks much for the initial work here, @victorsaccucci! The solution here was as follows:
- I removed all the cases where we were setting the title on tab switch, which your contribution turned me on to
- I set them all on view load for the installation tab
- The above ultimately works as we have the app set up to always load the installation tab first, as this is needed to assure that when new keyboards are installed, that the user cannot just navigate back to the settings tab which then would not load and update the available keyboards to changes settings for

Really great to have your support on the project!

Issue #540, opened 2025-06-24
When typing a special character in uppercase, it sets the character in lowercase


alejobarrera22:
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

This has been tested in spanish. 
When I try to type de word, for example "ÁMBITO", all characters in uppercase with the Uppercase Button locked. When locked, the keyboard types "áMBITO", this does not happen when you only click on the uppercase button once.
Step by step:
1. Given the user is typing on an app in spanish
2. When the user touches twice on the uppercase button 
3. Then the system locks the uppercase
4. When the user long-presses the "A" letter
5. Then the system shows all the options for the special A characters
6. When the user selects one of them, for example "Á"
7. Then the system types "á" in lowercase

Expected result: th system types "Á" in uppercase.

https://github.com/user-attachments/assets/e45d1572-7e1a-43d5-8201-f8827af783f3

### Device type

iPhone

### Versions

iOS 26 (Developer Beta), and Scribe 3.1.1

damien-rivet:
Hey ! Can I work on this issue?

andrewtavis:
Closed by #547 🚀 Thanks for another great contribution, @damien-rivet, and thanks so much for opening this issue, @alejobarrera22! Please do let us know if you see anything else 😊

PR #547, opened 2025-10-13
[#540] Fix caps locked capitalisation not taken into account

damien-rivet:
### Contributor checklist

---

### Description

When using the CapsLock shift button, the uppercase characters were not displayed correctly AND when tapping on them, they would always output the lowercase version of them.

A long time ago, during another contribution, I noticed that we were using 2 separate enums to handle the state of both the shift & capslock buttons, leading to complex conditions blocks throughout the code.

In this PR, I reconcile them under a single enum & single app property to simplify the management of those buttons.

This allowed me to fix both issue with a single stone.

### Screenshots

https://github.com/user-attachments/assets/b4ea1219-5d8d-4418-a40e-99801502127b

### Related issue

- #540

### DIFF
--- a/Keyboards/KeyboardsBase/KeyboardViewController.swift
+++ b/Keyboards/KeyboardsBase/KeyboardViewController.swift
@@ -2766,7 +2766,6 @@ class KeyboardViewController: UIInputViewController {
 
     case "shift":
       if shiftButtonState == .capsLocked {
-        // Return capitalization to default.
         shiftButtonState = .normal
       } else {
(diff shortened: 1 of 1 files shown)


andrewtavis (approved):
Tests are passing and the functionality is exactly what we're looking for 😊 Always great  to get contributions from you, @damien-rivet! Perfect timing as the changes for `pr_ci.yaml` will be integrated into other PRs that will make it easier on the Outreachy applicants we have in the community to contribute 😊

PR #553, opened 2025-12-02
Add word-for-word delete function

### LINKED ISSUE #301: Option for word by word deletion on delete long hold
### Terms

- [X] I have searched [open and closed feature requests](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue+label%3Afeature)

### Description

This issue would focus on creating a menu option that allows the user to enable word by word deletion rather than a consistent delete that speeds up as is now the case. I would say that this should be disabled by default, and this issue is further blocked by #16 as we need the menu before we can add in this option. I'll add it to the designs right now though 😊

### Contribution

Happy to support someone on this when we're done with #16, or will get to it myself once higher priority issues are finished :)

henrikth93:
This commit will add word-for-word delete function if you hold the delete button.

### Contributor checklist

- [] I have tested my code with the `xcodebuild` and `swiftlint --strict` commands as directed in the [testing section of the contributing guide](https://github.com/scribe-org/Scribe-iOS/blob/main/CONTRIBUTING.md#testing)

---

### Related issue

- Closes #301

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,8 @@ Emojis for the following are chosen based on [gitmoji](https://gitmoji.dev/).
 - Adds an English keyboard ([#7](https://github.com/scribe-org/Scribe-iOS/issues/7)).
 
 ### ✨ New Features
-- Users can now delete text word for word instead of character by character by enabling a function in the settings menu. The words are deleted by holding down backspace.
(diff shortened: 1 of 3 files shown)


andrewtavis (approved):
praise: Is working very well, @henrikth93! The words are being deleted word by word instantly, but I think that this is good enough to close the issue, and we can adjust if there's interest later on 😊

Issue #554, opened 2025-12-09
UI becomes unresponsive when switching Translation language


catreedle:
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

When changing the translation source language in the Translation Language settings, the app becomes unresponsive.

 Steps to reproduce
1. Open the app
2. Go to Settings
3. Select a language under Select installed keyboard
4. Open Translation language to choose a language

Result: App becomes unresponsive
Expected: Selected translation language is applied and app remains responsive

### Device type

iPhone

### Versions

iOS 18.6

PR #556, opened 2025-12-09
Hide unsupported languages from translation selection

catreedle:
### Contributor checklist

---

### Description

This PR hides languages that are not yet supported for translation from the translation selection list and prevents the UI from becoming unresponsive when selecting a translation language.
### Related issue

- #554


andrewtavis (approved):
praise: All's looking good here, @catreedle :) Thanks so much for the issue and for the quick fix. We can uncomment this later as we add in the languages 😊

andrewtavis:
note: I reset the changes to the XCode strings. I assume that that was from you running the emulator and then that editing the file. We should try to get to a point where the file that's generated in Scribe-i18n is appropriate for the emulator and thus won't be edited unless there are changes in Scribe-i18n. I'll make an issue 😊

PR #574, opened 2026-01-04
Fix crash in Settings when Indonesian keyboard is installed

### LINKED ISSUE #573: Crash in Settings with Indonesian keyboard installed — Fatal error: Abbreviation not found for language: Indonesian
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

When an Indonesian keyboard is installed on the device, opening the Settings screen causes the app to crash immediately.

The crash occurs with the following error:
“Fatal error: Abbreviation not found for language: Indonesian”

Expected behavior:
The app should handle the Indonesian keyboard safely and either map it to a valid abbreviation or fall back gracefully without crashing.

### Device type

iPhone

### Versions

Scribe-iOS: latest main branch

prince-0408:
Fixes #573

This PR prevents a crash in Settings when the Indonesian keyboard is installed by safely handling missing language abbreviations.


andrewtavis (approved):
praise: Thanks for the PR, @prince-0408! I think that the best thing to do for now is to disable the Indonesian keyboard as has been done with Danish and Norwegian. We can enable it when we start working on it again 😊

CC @catreedle for the change here 👋

PR #576, opened 2026-01-05
fix i18n keys word for word deletion

catreedle:
### Contributor checklist

---

### Description

Fix i18n keys word for word deletion and description in Scribe/SettingsTab/SettingsTableData.swift

### Related issue

- #ISSUE_NUMBER


andrewtavis:
@catreedle, it's currently looking like this :) Would we want to allow the menu label to be split to the next line? Might be a good idea in case other languages are longer than English.

Simulator Screenshot - iPhone 16e - 2026-01-06 at 00 19 52

catreedle:
Simulator Screenshot - iPhone 16e - 2026-01-06 at 13 45 58

hi @andrewtavis I tried fixing this and had to increase the row height in general. Hope it's okay.

andrewtavis:
The change for the radio button looks really good, @catreedle, but would we be able to reduce the spacing if there's a sub menu? See below:

Simulator Screenshot - iPhone 16e - 2026-01-06 at 10 46 56

I think it'd be great if we could reduce the above distance by half 🤔

catreedle:
> The change for the radio button looks really good, @catreedle, but would we be able to reduce the spacing if there's a sub menu? See below:
> 
> Simulator Screenshot - iPhone 16e - 2026-01-06 at 10 46 56
> I think it'd be great if we could reduce the above distance by half 🤔

I will look into it 😊

andrewtavis:
Merged in main so getting this in will be easier, @catreedle :)

andrewtavis (approved):
praise: So great to have this finalized and to also have the fix for the line height, @catreedle! Thanks for your hard work and the care that you put into the styling here 🎨

note: I sent an incorrect commit to your Scribe-i18n PR that was just merged, but realized it here and sent a fix along. Didn't realize that there's no space added at the end of the xcstrings file. Let's keep this in mind as we work, and maybe we need to remove `xcstrings_file.write("\n")` at the end of `convert_jsons_to_xcstrings.py`. Let me know what your thoughts are on that :)

PR #577, opened 2026-01-05
Fix tab bar appearance on iOS < 26

### LINKED ISSUE #571: App tab bar appearance should be reactive to iOS <26
### Terms

- [x] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [x] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

In #552 we did a fix for the liquid glass tab bar, but this introduced a bug. We need the tab bar to not be transparent if the iOS version of the device is <26. We also would like the transparency of the tab bar to be on all three tabs of the device, so `Installation`, `Settings` and `About`.

- Note: The changes in #560 should roughly be applied to all screens and these changes should be made optional based on the version :)

### Device type

iPhone and iPad

### Versions

Current main

prince-0408:
Fixes #571

Updates the tab bar appearance logic in AppDelegate to ensure correct behavior on iOS versions below 26.

Screenshot 2026-01-05 at 11 31 06 AM

### DIFF
--- a/Scribe/AppDelegate.swift
+++ b/Scribe/AppDelegate.swift
@@ -27,9 +27,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
       UITabBar.appearance().standardAppearance = appearance
       UITabBar.appearance().scrollEdgeAppearance = appearance
     } else {
-      // Fallback for earlier versions: non-transparent
       UITabBar.appearance().isTranslucent = false
-      UITabBar.appearance().barTintColor = .white // Default non-transparent color
+      UITabBar.appearance().barTintColor = .white // default non-transparent color
     }
 
     return true


Scribe/AppDelegate.swift
```diff
@@ -16,6 +16,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
     // Override point for customization after application launch.
+    if #available(iOS 13.0, *) {
```
andrewtavis: question: Why are we targeting iOS 13.0 for this. Shouldn't it be iOS 26?

Scribe/AppDelegate.swift
```diff
@@ -16,6 +16,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+      let appearance = UITabBarAppearance()
+      appearance.configureWithOpaqueBackground()
+      appearance.backgroundColor = .white
```
andrewtavis: note: For dark mode we have a different color for the tab bar. Based on the Figma designs, it should be white or black.

prince-0408: Okay I'll work on that

Scribe/AppDelegate.swift
prince-0408: On iOS 26 and newer, the system already applies the correct appearance automatically, so this code effectively becomes a no-op there.
On iOS 13–18, it ensures consistent, non-transparent tab bar behavior

andrewtavis (approved):
praise: All's working great here now, @prince-0408 😊 Thanks for the quick work and reactions to the review! :) :)

PR #579, opened 2026-01-08
Feature/colon to emoji toggle

### LINKED ISSUE #364: Discussion: `:` to emoji entry command for Scribe keyboards
### Terms

- [X] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue)

### Issue

This issue is to discuss whether Scribe should include a colon (`:`) to emoji entry command. Allowing people to use this in any app would be a cool feature 😊 We could have an option for it in the menu be auto-activated, and then use all the emoji data that we get to then replace the command keys with pattern matched options based on what the user types after `:`.

jatinfoujdar:
### Contributor checklist

- [x] This pull request is on a separate branch and not the main branch
- [x] I have tested my code with the `xcodebuild` and `swiftlint --strict` commands

---

### Description

This PR adds a new "Colon to emoji entry" toggle to the keyboard settings. When enabled, users can type a colon (:) followed by a keyword to get emoji suggestions, similar to Slack or Discord. If the toggle is off, this feature is disabled.

Main changes:

- Added the "Colon to emoji entry" toggle to the keyboard functionality section in settings.
- Updated the settings UI and logic to save and load the toggle state.
- Modified the keyboard logic so colon-to-emoji suggestions only appear if the toggle is enabled.
- Updated related models and variables to support this feature.

### Related issue
- Fixes #364

### DIFF
--- a/Keyboards/KeyboardsBase/Keyboard.xib
+++ b/Keyboards/KeyboardsBase/Keyboard.xib
@@ -60,10 +60,14 @@
                 <outlet property="padEmojiKey4" destination="d-PaEK-4" id="c-PaEK-4"/>
                 <outlet property="padEmojiKey5" destination="d-PaEK-5" id="c-PaEK-5"/>
                 <outlet property="phoneEmojiDivider" destination="6DZ-gU-SO5" id="VXK-ZI-d4Q"/>
+                <outlet property="phoneEmojiDivider1" destination="d-PED-1" id="c-PED-1"/>
+                <outlet property="phoneEmojiDivider2" destination="d-PED-2" id="c-PED-2"/>
                 <outlet property="phoneEmojiKey0" destination="pTv-BK-rKh" id="U16-av-iWx"/>
                 <outlet property="phoneEmojiKey1" destination="KVC-Ai-vdF" id="iYf-s9-OKK"/>
                 <outlet property="phoneEmojiKey2" destination="d-PEK-2" id="c-PEK-2"/>
-                <outlet property="phoneEmojiDivider1" destination="d-PED-1" id="c-PED-1"/>
+                <outlet property="phoneEmojiKey3" destination="d-PEK-3" id="c-PEK-3"/>
+                <outlet property="phoneEmojiKey4" destination="d-PEK-4" id="c-PEK-4"/>
+                <outlet property="phoneEmojiKey5" destination="d-PEK-5" id="c-PEK-5"/>
                 <outlet property="pluralKey" destination="FgS-T1-2rI" id="NSp-W5-nas"/>
                 <outlet property="rightAutoPartition" destination="hPX-B4-WMk" id="vhl-ch-h5H"/>
                 <outlet property="scribeKey" destination="MCB-7F-dNd" id="AzX-VF-3Nj"/>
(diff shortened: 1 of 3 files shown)


jatinfoujdar:
Hey @andrewtavis! 👋

I looked into why you might not have been seeing the emoji suggestions. It turns out there was a bug where the 'colon to emoji' feature (and a few other settings) were defaulting to false in UserDefaults if they hadn't been explicitly toggled in the app's settings yet.

I've just pushed a fix that ensures these features default to true, matching the intended behavior and the Settings UI.

To use it, simply type a : followed by a keyword (e.g., :smile) on any Scribe keyboard, and you should see the emoji suggestions appear in the top bar. Typing just the : should also immediately show the most common emojis.

Let me know if it works for you now! If you'd still like a video, I'd be happy to record one from the simulator. Thanks!

andrewtavis:
Thanks for the quick fix, @jatinfoujdar! I think I'm generally seeing what's intended here :) I guess now that I'm seeing it, there is an improvement upon the already existing emoji autosuggestion functionality, but maybe we can do more? 🤔

The `:` functionality on GitHub and other apps functions in a similar way, but the big plus is that there are more options than what we provide (two for phones and three for pads). Could `:` be used in a way that we could potentially show the user more emoji options? As of now we still have the two word autosuggestion fields and are just using the two emoji suggestions.

Maybe something that could be done is we could expand the emoji fields to be six if the user types `:`? This would require us to regenerate the data though as we just have three per word right now. Can be done though.

Another thing that I'm seeing is that there's a background behind the autosuggestions if the user has types `:`. See below:

Simulator Screenshot - iPhone 16e - 2026-01-11 at 20 01 48

Could we look into fixing the above, and from there it'd be great to get your feedback on the idea of expanding the number of emojis. I just think that the way that it is now, we're still showing autosuggested words and the UI looks a bit awkward. We could expand the number of emojis per word with Scribe-Data and then provide a full row of emojis that are updating as the user types :)

jatinfoujdar:
@andrewtavis  Thanks for the detailed feedback — I agree 🙂

The : entry is an improvement, but it’s still constrained by the word autosuggestion layout, which makes the UI feel a bit awkward right now.

I’ll look into fixing the autosuggestion background when typing : and hiding word suggestions when we’re clearly in emoji mode.

I also like the idea of expanding emoji suggestions to a full row (5–6) when : is active. That would likely require regenerating data in Scribe-Data and some layout changes, so it’ll take a bit of time to do properly, but it feels like the right direction.

Thanks again — happy to iterate on this 👍

andrewtavis:
Great to get your thoughts on this, @jatinfoujdar! Let's maybe start with the minor fixes to this PR so that we don't have the backgrounds being added to the suggestions. From there we can look into:

- Adding the other four emoji buttons to the layout
- Setting it up so that we have the left most emoji button being populated with the first emoji suggestion
- We can then expand the selection from the database to allow more than three options per word

Really looking forward to seeing how this progresses! Thanks for your efforts so far! 😊

andrewtavis:
Do we now have the extra emoji buttons at this point, @jatinfoujdar, or do we still need to add them? I.e. do we have six emoji buttons for phones and maybe nine for tablets?

jatinfoujdar:
@andrewtavis  I think 9 is a perfect limit for tablets! It strikes a great balance—any more would definitely feel too cluttered, but 9 makes excellent use of the extra screen space.

andrewtavis:
note, @jatinfoujdar: Check the prek directions that I added to this PR in ac10d6a :) This will warn you of the linting errors that came up that I fixed in 74328c8 😊

andrewtavis:
Fixes were just extra spacing :)

Issue #582, opened 2026-01-13
Investigate how to maintain conjugation application within Scribe-iOS


andrewtavis:
### Terms

- [x] I have searched [open and closed issues](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aissue)
- [x] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Issue

This issue will investigate how to maintain the Scribe-Conjugate iOS application in this repository. We used to have a separate repo for this, but the decision was made that we'd maintain both applications in one repository in the same way that Mozilla did/does for the mobile Firefox apps ([Android](https://github.com/mozilla-mobile/firefox-android) and [iOS](https://github.com/mozilla-mobile/firefox-ios)) and Bitwarden does for the password manager and authenticator ([Android](https://github.com/bitwarden/android) and [iOS](https://github.com/bitwarden/ios)).

This issue would specifically look at:
- [] How to maintain two Xcode projects in a single repo
- [] How to share files between the two projects (we want to be able to reuse most of the menu)
- [] Initial setup steps
- [] Updating the documentation (readme and contributing guide)

The first two points above can be recorded in the issue as a comment :)

Happy to support here as I can and review!

PR #586, opened 2026-01-20
feat:add a new storyboard, info.plist and target

angrezichatterbox:
### Contributor checklist

---

### Description

The PR adds the conjugate application and the necessary things needed to run an additional target. I have currently copied the existing storyboard of the app to be displayed for the conjugate app.

### Related issue

- #582


andrewtavis:
Thanks for opening the PR, btw! Amazing that we have progress on both projects now 😊

angrezichatterbox:
> Thanks for looking into the error, @angrezichatterbox! What ended up being the issue with that error? Would be good to discuss in the PR so we can reference the solution later 😊

The commit was no longer present was what I understood so I removed the submodule and added it again with the current one. I hope that doesn't break anything.

Issue #510, opened 2024-10-02
[iPad] Same key appearing at different positions on same keyboard can have wrong popup


christianbilodeau:
### Terms

- [X] I have searched all [open bug reports](https://github.com/scribe-org/Scribe-iOS/issues?q=is%3Aopen+is%3Aissue+label%3Abug)
- [X] I agree to follow Scribe-iOS' [Code of Conduct](https://github.com/scribe-org/Scribe-iOS/blob/main/.github/CODE_OF_CONDUCT.md)

### Behavior

On larger iPad models, the same key can appear twice in the same layout. The key objects are identified with their value, so there is no current way to differentiate two keys having the same character but appearing at different positions. This can result in having the wrong popup when pressing the key.

An example is the swedish keyboard for the character §:

Correct popup (is left):

<img width="573" alt="Screenshot 2024-10-02 at 09 54 55" src="https://github.com/user-attachments/assets/848c6da7-9094-4b9e-9b23-a4315bef79bf">

Wrong popup (should be right but is identified as the left key):

<img width="564" alt="Screenshot 2024-10-02 at 09 56 56" src="https://github.com/user-attachments/assets/6a75e8bb-7a9c-40d6-8b60-5bd67ba0a4fe">

### Device type

iPad

### Versions

iOS 18, Scribe 3.1.1

PR #632, opened 2026-03-29
Add index to key animation

henrikth93:
This commit will add an index variable so we can differentiate between keys that share the same character.

### Contributor checklist

- [] I have tested my code with the `xcodebuild` and `swiftlint --strict` commands as directed in the [testing section of the contributing guide](https://github.com/scribe-org/Scribe-iOS/blob/main/CONTRIBUTING.md#testing)

---

### Related issue

- #510

### DIFF
--- a/Keyboards/KeyboardsBase/InterfaceVariables.swift
+++ b/Keyboards/KeyboardsBase/InterfaceVariables.swift
@@ -246,13 +246,6 @@ func setKeyboardLayout() {
         pluralPromptAndColorPlaceholder
     ]
 }
-
-// Variables that define which keys are positioned on the very left, right or in the center of the keyboard.
-// The purpose of these is to define which key pop up functions should be ran.
-var centralKeyChars = [String]()
-var leftKeyChars = [String]()
-var rightKeyChars = [String]()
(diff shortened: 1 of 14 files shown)


henrikth93:
> Are we ready for a final review here, @henrikth93? :)

Yes!

EDIT: One thing I thought of is that maybe I should not use magic numbers, and instead use variables?

andrewtavis:
Typing out a discussion here from the sync - for this in particular:

- We want this to be row indexed based, so index 0 within a row would be left most key and index -1 within a row would be the right most
- We would need to have conditions based on if there's a space to the left and right of the row - i.e. keyboards like Portuguese and English in the middle row have extra spaces and would thus get a middle popup

henrikth93:
> Typing out a discussion here from the sync - for this in particular:
> 
> * We want this to be row indexed based, so index 0 within a row would be left most key and index -1 within a row would be the right most
> * We would need to have conditions based on if there's a space to the left and right of the row - i.e. keyboards like Portuguese and English in the middle row have extra spaces and would thus get a middle popup

Ready for a review

